PR #2, opened 2019-07-31
Int2019-11 registration frontend

robertpin:


ClientApp/viking-vault/src/App.tsx
```diff
@@ -1,24 +1,12 @@
 import React from 'react';
-import logo from './logo.svg';
+// import logo from './logo.svg';
```
vasi-axinte: Please remove commented code.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
+import React from "react";
+import {ResponseModal} from "./ReponseModal";
+import {Redirect, BrowserRouter as Router} from "react-router-dom";
+
+const bcrypt = require("bcryptjs");
+const url = "https://localhost:44323/api/";
```
vasi-axinte: This should be renamed to "baseUrl" since that is what it represents.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+            method: "POST",
+            headers: {
+              'Accept': 'application/json',
+              'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({
+              email: this.state.email,
+            })
+        },).then(response => response.json()).then(result => {
+            if(result === "unique") {
```
vasi-axinte: This should be changed to receive a boolean value from backend. Strings are prone to errors.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+                })
+            }
+            else if(result === "not-unique") {
+                this.setState({
+                    valid: "invalid"
+                })
+            }
+            else {
+                this.setState({
+                    valid: "null"
```
vasi-axinte: How would you explain that the validation status is "null" for something?

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+                })
+            }
+        });
+    }
+
+    private handleChange = (val: string, inpName: string) => {
+        this.setState({
+            [inpName]: val
```
vasi-axinte: Please don`t use abbreviations in your code. inp and val can mean different things for different people.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+        })
+    }
+
+    private handleEmailChange = (e: any) => {
+        this.setState({
+            email: e.target.value
+        }, this.validateEmail);
+    }
+
+    private emailError = () => {
```
vasi-axinte: All method names should be verbs. This is a noun that doesn`t suggest what the method does.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+        if(!emailRegEx.test(this.state.email)) {
+            return  {
+                msg: "Email format incorrect",
```
vasi-axinte: Please don`t use abbreviations. msg could mean something else for a different person.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+                class: "alert alert-danger"
+            }
+        }
+        if(this.state.valid === "invalid") {
```
vasi-axinte: Use booleans instead of strings.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+                class: "alert alert-success"
+            }
+        }
+        return  {
+            msg: "Server error. Try again later",
+            class: "alert alert-danger"
+        }
+    }
+
+    private handleCPasswordChange = (e: any) => {
```
vasi-axinte: I think that there is a typo here. If not, than what does CPassword stands for? Please do not abbreviate.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+        this.setState({
+            cpassword: e.target.value
+        }, this.passwordMatch)
+    }
+
+    private passwordMatch = () => {
```
vasi-axinte: Please use verbs for all the methods.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+            val = false;
+        }
+        if(this.state.address === "" || this.state.cnp === "" || this.state.firstName === "" 
+            || this.state.lastName === "" || this.state.password === "") {
+                val = false;
+        }
+        return val;
+    }
+
+    private execute = async () => {
```
vasi-axinte: Uhm... execute what? Also, method is too long. Try to refactor the code such that all the method have maximum 15 lines.

ClientApp/viking-vault/src/ReponseModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+class ResponseModal extends React.Component<any, any> {
+    constructor(props: any) {
+        super(props);
+        this.state = {
+            open: props.open
+        }
+    }
+
+    static getDerivedStateFromProps(props: any, state: any) {
```
vasi-axinte: Why is this method static?

ClientApp/viking-vault/src/ReponseModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+        if(props.open !== state.open)
+            return {
+                open: props.open
+            }
+        return null;
+    }
+
+    private closeModal = () => {
+        // console.log("trying to close modal");
```
vasi-axinte: Please remove commented code.

ClientApp/viking-vault/src/ReponseModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+        this.props.modalClose(false);
+        this.setState({
+            open: false
+        });
+    }
+
+    render() {
+        return (<div className="container">
+            <div className="modal" id="myModal" role="dialog" style={this.state.open?modalOpenStyle: modalCloseStyle}>
```
vasi-axinte: myModal? Hmmm. Also, please use css classes instead of style.

Server/VikingVault.API/Controllers/LoginController.cs
```diff
@@ -0,0 +1,33 @@
...
+        public LoginController(ILoginService loginService)
+        {
+            _loginService = loginService;
+        }
+
+        [AllowAnonymous]
+        [HttpPost("authenticate")]
+        public IActionResult Authenticate([FromBody]UserLogin userParam)
+        {
+            var user = _loginService.Authenticate(userParam.Email, userParam.Password);
```
vasi-axinte: userParam can simply be renamed to user or userLoginData

Server/VikingVault.API/Controllers/UniqueEmailController.cs
```diff
@@ -0,0 +1,59 @@
...
+using VikingVault.Services.Abstractions;
+using VikingVault.DataAccess.Models;
+
+namespace VikingVault.API.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class UniqueEmailController : ControllerBase
+    {
+
```
vasi-axinte: Please remove these empty lines.

Server/VikingVault.API/Controllers/UniqueEmailController.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly IUniqueEmailService _uniqueEmailService;
+
+        public UniqueEmailController(IUniqueEmailService uniqueEmailService)
+        {
+            _uniqueEmailService = uniqueEmailService;
+        }
+        
+        [HttpGet]
+        public ActionResult Get()
```
vasi-axinte: Do we still need this?

Server/VikingVault.API/Controllers/UniqueEmailController.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            return Ok("good job");
+        }
+
+
+        // POST: api/UniqueEmail
```
vasi-axinte: Please remove this comment.

vasi-axinte (changes requested):

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,233 @@
+import React from "react";
+import {ResponseModal} from "./ReponseModal";
+import {Redirect, BrowserRouter as Router} from "react-router-dom";
+
+const url = "https://localhost:44323/api/";
+const emailRegEx = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+interface IFormState {
+    [key: string]: string | number | boolean;
+    email: string;
```
Slitthe, line 12: It's good to put related object properties into their own object, it seems to me like email, password, cpassword, firstName...etc are all part of the same form and represent the same entity.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,233 @@
...
+    openModal: boolean;
+    redirect: boolean;
+}
+
+class RegisterForm extends React.Component<any, IFormState> {
+    constructor(props: any) {
+        super(props);
+
+        this.state = {
+            email: "",
```
Slitthe: Same here as the interface.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,233 @@
+import React from "react";
+import {ResponseModal} from "./ReponseModal";
+import {Redirect, BrowserRouter as Router} from "react-router-dom";
+
+const url = "https://localhost:44323/api/";
```
Slitthe: Since this is used in other areas of the code, its better to make a file of constants and use the same constant all throughout the application, it prevents a lot of bugs and when you deploy it in za cloud you will need to change the api endpoint everywhere.

ClientApp/viking-vault/src/ReponseModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+    display: "block",
+    opacity: 1,
+    backgroundColor: "rgba(0,0,0,0.5)"
+}
+
+const modalCloseStyle = {
+    display: "none"
+}
+
+class ResponseModal extends React.Component<any, any> {
```
Slitthe: `<any, any>` Please use types whenever you can for state and props, I know any is easy to use when debugging but it helps a lot to know later on when you come back to look at the component 2 weeks later what it takes as an input and what data it holds in its state. Also a big advantage is that when you use the component externally you get props autocomplete.

ClientApp/viking-vault/src/ReponseModal.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+
+const modalOpenStyle = {
```
Slitthe: inline styles are generally best used when you actually need to compute values using JavaScript rather than using values which do not change.

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,233 @@
...
+                    openModal: true
+                });
+            }
+            return response.json();
+        }).then(result => {
+            setTimeout(() => {
+                this.setState({
+                    redirect: true
+                });
+            }, 1500);
```
Slitthe, line 209: Are you adding an artificial delay to the redirect?

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,233 @@
...
+        return val;
+    }
+
+    private execute = async () => {
+        this.setState({
+            openModal: true,
+            response: "Please wait..."
+        });
+        const user = {
+            Email: this.state.email,
```
Slitthe: naming conventions for JavaScript object keys is camelCase,

ClientApp/viking-vault/src/RegisterForm.tsx
robertpin: Yes :D

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,233 @@
...
+            || this.state.lastName === "" || this.state.password === "") {
+                val = false;
+        }
+        return val;
+    }
+
+    private execute = async () => {
+        this.setState({
+            openModal: true,
+            response: "Please wait..."
```
Slitthe, line 166: Would be nice to add a spinner instead of `'Please wait...`, not really critical issue but it helps with the User Experience and it looks pretty

ClientApp/viking-vault/src/RegisterForm.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+    private mandatoryFieldsCompletedCorrectly = () => {
+        let val = true;
+        if(this.emailError().msg !== "Valid") {
+            val = false;
+        }
+        if(this.passwordMatch().msg !== "Passwords match") {
+            val = false;
+        }
+        if(this.state.address === "" || this.state.cnp === "" || this.state.firstName === ""
```
Slitthe: You could refactor this using the Array method of .every(), like `["address", "cnp", "firstName", "lastName"].every(keyName => this.state[keyName] !== "")`

ClientApp/viking-vault/src/RegisterForm.tsx
robertpin: I can't because state.valid can have 3 values: valid, invalid and error.

ClientApp/viking-vault/src/ReponseModal.tsx
robertpin: It's like a lifecycle hook. 
https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

ClientApp/viking-vault/src/RegisterForm.tsx
robertpin: I'm not sure I understand. It's not ok that all of them are in the same interface?

Slitthe: What I mean is, that if you have related properties, let's say that you have a person and some other state properties 


```
interface IFormState {
	[key: sring]: number;
	isModalOpen: boolean;
	isLoading: boolean;
	otherComponentRelatedState: SoemOtherType;
	firstName: string;
	lastName: string;
	age: number;
}
```

you'd better have something like this 

```
interface IFormState {
	[key: sring]: number;
	isModalOpen: boolean;
	isLoading: boolean;
	otherComponentRelatedState: SoemOtherType;
	person: {
		firstName: string;
		lastName: string;
		age: number;
	}
	
}
```


or even better


```
interface IPerson {
	firstName: string;
	lastName: string;
	age: number;
}

interface IFormState {
	[key: sring]: number;
	isModalOpen: boolean;
	isLoading: boolean;
	otherComponentRelatedState: SoemOtherType;
	person: IPerson	
}
```

ClientApp/viking-vault/src/RegisterForm.tsx
robertpin: ...

ClientApp/viking-vault/src/RegisterForm.tsx
robertpin: .. It's kinda weird to refactor it. I mean it basically creates an user object and sends it to the backend and then shows the response. The fetch alone has 15 lines of code and I can't exactly break it into smaller pieces. It's one function call

ClientApp/viking-vault/src/RegisterForm.tsx
robertpin: Ok. I understand now. Thank you!

ClientApp/viking-vault/src/App.css
```diff
@@ -1,33 +1,54 @@
...
 
-@keyframes App-logo-spin {
-  from {
-    transform: rotate(0deg);
-  }
-  to {
-    transform: rotate(360deg);
-  }
+.descriptionText {
+  /* color:white;
```
vasi-axinte: Please remove commented code.

ClientApp/viking-vault/src/RegisterForm.tsx
vasi-axinte: Then use nullable booleans or some enums. Strings are a bad choice for this kind of thing since they can be easily misspelled.

ClientApp/viking-vault/src/RegisterForm.tsx
vasi-axinte: From what you are saying, there are many things that can be separated in standalone methods. For example, everything you`ve said between "and" can be a method. (i.e. sendData, showResponse etc)

ClientApp/viking-vault/src/RegisterForm.tsx
vasi-axinte: Why?

vasi-axinte (changes requested):

ClientApp/viking-vault/src/RegisterForm.tsx
robertpin: To be done in a future sprint if Product Owner says it's necessary?

vasi-axinte (approved):

PR #8, opened 2019-08-06
Made images into links. Created admin controller on backend to check …

robertpin:
…if user is admin

### DIFF
--- a/ClientApp/viking-vault/src/App.tsx
+++ b/ClientApp/viking-vault/src/App.tsx
@@ -1,20 +1,10 @@
 import React from 'react';
-import { ProfilePage } from './ProfilePage';
-import './App.css';
-import { BrowserRouter as Router, Route, Link } from "react-router-dom";
-import { LoginForm } from './Login'
-import { RegisterForm } from "./RegisterForm";
-import "./ProfilePage.css"
-import UserPage from './components/UserPage';
+import { Router } from './Router';
 
 const App: React.FC = () => {
   return (
     <div className="App">
-      <Router>
-        <Route path="/register/" exact component={RegisterForm} />
-        <Route path="/" exact component={LoginForm}/>
-        <Route path="/user" component={UserPage} />
-      </Router>
+      <Router></Router>
     </div>
   );
 }
--- a/ClientApp/viking-vault/src/App.tsx
+++ b/ClientApp/viking-vault/src/App.tsx
@@ -12,7 +12,7 @@ const App: React.FC = () => {
     <div className="App">
       <Router>
         <Route path="/register/" exact component={RegisterForm} />
-        <Route path="/login/" exact component={LoginForm}/>
+        <Route path="/" exact component={LoginForm}/>
         <Route path="/user" component={UserPage} />
       </Router>
     </div>
--- a/ClientApp/viking-vault/src/Router.tsx
+++ b/ClientApp/viking-vault/src/Router.tsx
@@ -31,16 +31,9 @@ class Router extends React.Component<any, IState> {
             }
         }).then(response => response.json())
         .then(result => {
-            if(result === true) {
-                this.setState({
-                    isAdmin: true
-                })
-            }
-            else {
-                this.setState({
-                    isAdmin: false
-                })
-            }
+            this.setState({
+                isAdmin: result
+            })
         });
     }
 
@@ -48,16 +41,18 @@ class Router extends React.Component<any, IState> {
         this.isUserAdmin();
     }
 
+    private makeRedirect() {
+        if(sessionStorage.getItem("Authentication-Token") === null)
+            return <Redirect to="/login" />;
+        else
+            return <Redirect to={this.state.isAdmin? "/admin" : "/user"} />;
+    }
+
     render() {
         return <BRouter>
             <Route path="/register/" exact component={RegisterForm} />
             <Route path="/login" exact component={LoginForm} />
-            <Route path="/" exact render={() => {
-                if(sessionStorage.getItem("Authentication-Token") === null)
-                    return <Redirect to="/login" />;
-                else
-                    return <Redirect to={this.state.isAdmin? "/admin" : "/user"} />;
-            }}/>
+            <Route path="/" exact render={() => this.makeRedirect()}/>
             <Route path="/user" component={UserPage} />
         </BRouter>
     }
--- a/ClientApp/viking-vault/src/Router.tsx
+++ b/ClientApp/viking-vault/src/Router.tsx
@@ -1,0 +1,66 @@
+import React from "react";
+import { Route, Redirect, BrowserRouter as BRouter } from 'react-router-dom';
+import { RegisterForm } from "./RegisterForm";
+import { LoginForm } from "./Login";
+import UserPage from './components/UserPage';
+import {constants} from "./ConstantVariables";
+import "./App.css"
+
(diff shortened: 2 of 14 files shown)


Server/VikingVault.API/Controllers/AdminController.cs
```diff
@@ -0,0 +1,33 @@
...
+    public class AdminController : ControllerBase
+    {
+        private readonly IAdminService _adminService;
+        
+        public AdminController(IAdminService adminService)
+        {
+            _adminService = adminService;
+        }
+
+        [HttpPost]
```
vladbuda: Should this be a POST request? How about a GET request?

ClientApp/viking-vault/src/App.tsx
```diff
@@ -12,7 +12,7 @@ const App: React.FC = () => {
     <div className="App">
       <Router>
         <Route path="/register/" exact component={RegisterForm} />
-        <Route path="/login/" exact component={LoginForm}/>
+        <Route path="/" exact component={LoginForm}/>
```
vasi-axinte: Default path should take us to the dashboard, not login. If the user isn`t logged in, the router should be the one that redirects to the login page instead.

ClientApp/viking-vault/src/components/TopBar.tsx
```diff
@@ -1,13 +1,58 @@
...
+import {variables} from "../ConstantVariables";
 
-class TopBar extends React.Component{
+const baseUrl = variables.baseUrl;
+
+interface IState {
+    isAdmin: boolean;
+}
+
+class TopBar extends React.Component<any, IState>{
```
vasi-axinte, line 13: Please use types for state and props :)

Server/VikingVault.Services/AdminService.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            _dbContext = dbContext;
+        }
+
+        public bool IsAdmin(string token)
+        {
+            var tokenObject = new JwtSecurityToken(token);
+            string userId = tokenObject.Payload["Id"].ToString();
+            var returnedUser = _dbContext.User.SingleOrDefault(u => u.Id == Int32.Parse(userId));
+            if (returnedUser.Email == "admin")
```
vasi-axinte: can`t we directly return "returnedUser.Email == "admin" ?

vasi-axinte (changes requested):

ClientApp/viking-vault/src/App.tsx
robertpin: I changed the routes so that the Login page is now on / and not on /login to avoid homepage redirects. If a user comes to the page it shouldn't be redirected to /login

ClientApp/viking-vault/src/components/TopBar.tsx
robertpin: But the component has no props. We don't need props for this component

ClientApp/viking-vault/src/components/TopBar.tsx
```diff
@@ -1,13 +1,58 @@
 import React from 'react'
 import './styles.css'
 import logo from './images/logo.png'
+import { Link } from 'react-router-dom';
+import {variables} from "../ConstantVariables";
```
Slitthe: wouldn't it make more sense to name this `constants` instead of `variables`?

ClientApp/viking-vault/src/components/TopBar.tsx
```diff
@@ -1,13 +1,58 @@
 import React from 'react'
```
Slitthe, line 1: Small thing, but try to group imports in a way, usually when you import an npm module you put them at the top, then group images and css files, then other components and external objects/files

Server/VikingVault.Services/AdminService.cs
```diff
@@ -0,0 +1,30 @@
...
+        public AdminService(VikingVaultDbContext dbContext)
+        {
+            _dbContext = dbContext;
+        }
+
+        public bool IsAdmin(string token)
+        {
+            var tokenObject = new JwtSecurityToken(token);
+            string userId = tokenObject.Payload["Id"].ToString();
+            var returnedUser = _dbContext.User.SingleOrDefault(u => u.Id == Int32.Parse(userId));
```
Slitthe: When the return type is clear, it's okay to use `var`, for example in `var tokenObject = new JwtSecurityToken(token);`, but when it's not, like this `var returnedUser = _dbContext.User.SingleOrDefault(u => u.Id == Int32.Parse(userId));`, it's better to use an explicit type instead of `var`

Slitthe: Just a warning here, when you use the `.Parse()`, it can fail to parse, so keep that in mind. It's usually better to use `.TryParse()` as the default parsing method. You don't need to change anything here, just keep that in mind (you probably knew that already but reminders are good 😅 )

ClientApp/viking-vault/src/components/TopBar.tsx
```diff
@@ -1,13 +1,58 @@
...
+        fetch(baseUrl+"admin", {
+            method: "POST",
+            headers: {
+              'Accept': 'application/json',
+              'Content-Type': 'application/json',
+              'x-access-token' : token.toString()
+            }
+        }).then(response => response.json())
+        .then(result => {
+            if(result === true) {
```
Slitthe: Why do you need the if/else blocks here, couldn't you simple say `this.setState({ isAdmin = result });`

ClientApp/viking-vault/src/components/TopBar.tsx
```diff
@@ -1,13 +1,58 @@
...
+const baseUrl = variables.baseUrl;
+
+interface IState {
+    isAdmin: boolean;
+}
+
+class TopBar extends React.Component<any, IState>{
+    constructor(props:any) {
+        super(props);
+        this.state = {
```
Slitthe: For constructors that simply set a state, you can get rid of them and set the state as a class property
https://miro.medium.com/max/700/0*OKQ-h5NIfbiH8OQD.png

vasi-axinte (approved):

ClientApp/viking-vault/src/Router.tsx
```diff
@@ -0,0 +1,66 @@
...
+    componentDidMount() {
+        this.isUserAdmin();
+    }
+
+    render() {
+        return <BRouter>
+            <Route path="/register/" exact component={RegisterForm} />
+            <Route path="/login" exact component={LoginForm} />
+            <Route path="/" exact render={() => {
+                if(sessionStorage.getItem("Authentication-Token") === null)
```
Slitthe: Please extract this into a class method, usually it's not very good to have functions written inside of JSX, it makes the code harder to read, to understand and to debug

ClientApp/viking-vault/src/Router.tsx
```diff
@@ -0,0 +1,66 @@
...
+            method: "GET",
+            headers: {
+              'Accept': 'application/json',
+              'Content-Type': 'application/json',
+              'x-access-token' : token.toString()
+            }
+        }).then(response => response.json())
+        .then(result => {
+            if(result === true) {
+                this.setState({
```
Slitthe: Please also get rid of the if/else statements in here and replace them with `this.setState({ isAdmin: result })`

PR #10, opened 2019-08-06
Created bank account model and added 4 fixed bank accounts at each us…

vladbuda:
…er registration

### DIFF
--- a/Server/VikingVault.Services/BankAccountService.cs
+++ b/Server/VikingVault.Services/BankAccountService.cs
@@ -1,0 +1,35 @@
+﻿using Microsoft.EntityFrameworkCore;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using VikingVault.DataAccess;
+using VikingVault.DataAccess.Models;
+using VikingVault.Services.Abstractions;
+using VikingVault.Services.Exceptions;
+
+namespace VikingVault.Services
+{
+    public class BankAccountService : IBankAccountService
+    {
+        private readonly VikingVaultDbContext _dbContext;
+
+        public BankAccountService(VikingVaultDbContext dbContext)
+        {
+            _dbContext = dbContext;
+        }
+
+        public BankAccount createBankAccount(BankAccount account)
+        {
+            try
+            {
+                _dbContext.Add(account);
+                _dbContext.SaveChanges();
+            }
+            catch (Exception e)
+            {
+                if (e is DbUpdateException || e is DbUpdateConcurrencyException) throw new BankAccountServiceException();
+            }
+            return account;
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


robertpin (approved):

Server/VikingVault.Services/BankAccountService.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class BankAccountService : IBankAccountService
+    {
+        private readonly VikingVaultDbContext _dbContext;
+
+        public BankAccountService(VikingVaultDbContext dbContext)
+        {
+            _dbContext = dbContext;
+        }
+
+        public BankAccount createBankAccount(BankAccount account)
```
vasi-axinte: This method name doesn`t follow the C# naming convention. PascalCase should be used instead.

Server/VikingVault.Services/UserService.cs
```diff
@@ -13,23 +13,63 @@ namespace VikingVault.Services
...
             _dbContext = dbContext;
+            _bankAccountService = bankAccountService;
         }
 
         public User CreateUser(User user)
         {
             user.Role = "user";
+            user.Password = PasswordEncryption.ComputeSha256Hash(user.Password);
+
+            BankAccount ronBankAccount = new BankAccount
```
vasi-axinte: This chunks of code can be extracted to methods "CreateBankAccount" that get some parameters. That way, we would reuse some code and everything would be smoother and easier to read.

vasi-axinte (changes requested):

Server/VikingVault.Services/UserService.cs
```diff
@@ -13,23 +13,34 @@ namespace VikingVault.Services
...
         {
             user.Role = "user";
+            user.Password = PasswordEncryption.ComputeSha256Hash(user.Password);
+
             try
             {
-                user.Password = PasswordEncryption.ComputeSha256Hash(user.Password);
                 _dbContext.Add(user);
+
+                _bankAccountService.CreateBankAccount(this.CreateBankAccount(user, "ron"));
```
Slitthe: I don't know the application's business logic, but it seems like this "ron", "eur".. strings could be replaced with an enum

Server/VikingVault.Services/BankAccountService.cs
```diff
@@ -0,0 +1,35 @@
...
+        public BankAccount CreateBankAccount(BankAccount account)
+        {
+            try
+            {
+                _dbContext.Add(account);
+                _dbContext.SaveChanges();
+            }
+            catch (Exception e)
+            {
+                if (e is DbUpdateException || e is DbUpdateConcurrencyException) throw new BankAccountServiceException();
```
Slitthe: Please try to avoid using one-liners statements, even though it's more lines of code to say

if () 
{
}
, it's easier to read and is more consistent and less prone to errors

vasi-axinte (approved):

PR #11, opened 2019-08-06
Int2019 22 collapsed menu

moleguaco99:


ClientApp/viking-vault/src/components/styles.css
```diff
@@ -34,8 +34,10 @@
 }
 
 .regular-div{
```
vasi-axinte: regular-div is not a suggestive enough name.

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -34,8 +34,10 @@
...
+    padding-top: 15px;
+    padding-left: 15%;
     border-bottom: 1px solid #09424d;
-    height:46px;
+    height:60px;
 }
 
 .img{
```
vasi-axinte: Having a img class isn`t a best practice. The name should be more descriptive.

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -185,4 +197,9 @@ hr{
     display: inline-block;
     font-size: 16px;
     cursor: pointer;
+}
+
+.accountsTitle{
```
vasi-axinte: This name should respect the kebab-casing

vasi-axinte (changes requested):

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -7,7 +7,7 @@
     transition: 0.5s;
     position: absolute;
     box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
-    width: 150px;
+    width: 60px;
 }
 
 .sidebar-false{
```
Slitthe: Please avoid using this class naming of "base-" + "true"/"false", it's really confusing, you could for example use `.sidebar.expanded` or `sidebar-collapsed` and `sidebar-expanded` (if that's what this true/false means when referring to the sidebar)

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -131,21 +133,31 @@
...
+    display:none;
 }
 
-.paragraph-false{
+.span-false{
     color: #09424d;
     font-weight: lighter;
 }
+
+.dashboard-true{
```
Slitthe: Same as above regarding the CSS class naming of "base" + "false"/"true"

ClientApp/viking-vault/src/components/SideBar.tsx
```diff
@@ -34,25 +34,28 @@ class SideBar extends React.Component<any, ISideBarState> {
         return(
             <div className={"sidebar-" + this.state.show} id="mySidebar">
                 <div className="regular-div">
-                    <p className="p"> <img className="img" src={dashboard} alt=""/> My Dashboard  <img src={arrow} className = {this.state.show ? "transform-none img" : "transform-reverse img"} alt="" onClick={this.clickHandler}/></p>
+                        <img className="img" src={dashboard} alt=""/> 
+                        &nbsp;
+                        <span className = {"dashboard-" + this.state.show}>Dashboard</span>
+                        <img src={arrow} className = {this.state.show ? "transform-none img" : "transform-reverse img"} alt="" onClick={this.clickHandler}/>
```
Slitthe: Small thing, but if the alt attribute is empty then it's kind of redundant to be there

ClientApp/viking-vault/src/components/SideBar.tsx
```diff
@@ -34,25 +34,28 @@ class SideBar extends React.Component<any, ISideBarState> {
         return(
             <div className={"sidebar-" + this.state.show} id="mySidebar">
                 <div className="regular-div">
-                    <p className="p"> <img className="img" src={dashboard} alt=""/> My Dashboard  <img src={arrow} className = {this.state.show ? "transform-none img" : "transform-reverse img"} alt="" onClick={this.clickHandler}/></p>
```
Slitthe, line 37: Could you maybe give more suggestive names for the classes of `regular-div` and `p`?, for example you could name it `standard-container` if that's what it does..?

ClientApp/viking-vault/src/components/SideBar.tsx
```diff
@@ -34,25 +34,28 @@ class SideBar extends React.Component<any, ISideBarState> {
...
-                <p className = {"paragraph-" + this.state.show + " p"}> <img className="img" src={block} alt=""/> &nbsp; Block card </p>
+                        <img className="img" src={block} alt=""/> <span className = {"span-" + this.state.show}>Block card</span>
                 </div>
                 <div className = {"side-menu-" + this.state.show}>
-                <p className = {"paragraph-" + this.state.show + " p"}> <img className="img" src={placeholder} alt=""/> &nbsp; Locations </p>
+                        <img className="img" src={placeholder} alt=""/> <span className = {"span-" + this.state.show}>Locations</span>
                 </div>
                 <div className = {"side-menu-" + this.state.show}>
-                <p className = {"paragraph-" + this.state.show + " p"}> <img className="img" src={card} alt=""/> &nbsp; Contact </p>
+                        <img className="img" src={card} alt=""/> <span className = {"span-" + this.state.show}>Contact</span>
```
Slitthe: It seems to me that the expression of `"span-" + this.state.show` is used multiple times in the JSX; you could extract this into a variable at the top of the render and use that variable in the JSX elements


let foo: string = "span-" + this.state.show;
...
...
<img className={foo}

ClientApp/viking-vault/src/components/SideBar.tsx
moleguaco99: It's okay, I erased all those p's anyway :3

ClientApp/viking-vault/src/components/styles.css
moleguaco99: 'Tis was solved

ClientApp/viking-vault/src/components/SideBar.tsx
```diff
@@ -31,28 +33,37 @@ class SideBar extends React.Component<any, ISideBarState> {
...
+
+                <div className="dashboard-div">
+                        <img className="menu-icon" src={dashboard} /> 
+                        &nbsp;
+                        <span className = {dashboard_visibility}>Dashboard</span>
+                        <img src={arrow} className = {this.state.show ? "transform-none menu-icon" : "transform-reverse menu-icon"} alt="" onClick={this.clickHandler}/>
                 </div>
-                <div className = {"side-menu-" + this.state.show}>
-                <p className = {"paragraph-" + this.state.show + " p"}> <img className="img" src={transfer} alt=""/> &nbsp; Transfer </p>
+                <div className = {side_menu_visibility}>
```
vasi-axinte: That variable name should be camelCased

vasi-axinte (changes requested):

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -33,12 +29,14 @@
     vertical-align: middle;
 }
 
-.regular-div{
+.dashboard-div{
```
vasi-axinte: dashboard-div isn`t specific enough. More exactly, that -div has no real purpose.

vasi-axinte (changes requested):

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -131,23 +129,30 @@
 }
 
 .side-menu-true {
```
Slitthe: Please try to find a better way to express a boolean state that an HTML component can have rather than having "class-" + "true"/"false", this is applicable all throughout the application, not just here

ClientApp/viking-vault/src/components/SideBar.tsx
```diff
@@ -31,28 +33,37 @@ class SideBar extends React.Component<any, ISideBarState> {
     }
 
     render(){
+        
+        let sidebarVisibility:string = this.state.show ? "sidebar collapsed" : "sidebar expanded";
```
Slitthe: Best practice in JavaScript is that if that variable is not changed, you make it a `const`

vasi-axinte (approved):

PR #12, opened 2019-08-07
Int2019 31 user transaction list done

robertpin:


moleguaco99 (approved):

moleguaco99 (approved):
Everything is just fine

Server/VikingVault.Services/TransactionService.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public Transaction AddTransaction(Transaction tran)
+        {
+            throw new NotImplementedException();
+        }
+
+        public List<Transaction> GetTransactions(string userId)
+        {
+            int uid = int.Parse(userId);
+            return _dbContext.Transactions.Where(t => t.user.Id == uid).OrderByDescending(t => t.Date).Take(5).ToList();
```
Slitthe: It's better to move these kind of variables outside (the `Take(5)` one), instead of writing them inline

ClientApp/viking-vault/src/UserDashboard/TransactionList.css
```diff
@@ -0,0 +1,19 @@
...
+
+.transactions table {
+    color: #58595B;
+}
+
+.transactions h5 {
+    color: #004450;
+}
+
+.transactions img {
```
Slitthe: It's better to avoid using elements names as much as possible and instead to use classes, so if the image represent an user image instead of saying `.transactions img`, you should rather say `.transactions .user-image`

Server/VikingVault.API/Controllers/TransactionController.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace VikingVault.API.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class TransactionController : ControllerBase
+    {
+        private readonly ITransactionService _transactionService;
+
+        public TransactionController(ITransactionService ts)
```
vasi-axinte: What is ts?

vasi-axinte (changes requested):

vasi-axinte (approved):

PR #14, opened 2019-08-08
Int2019 30 user account page

moleguaco99:


ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -19,18 +26,39 @@ class AccountPage extends React.Component<any, IAccountState>{
         super(props);
```
ed2298: Please remove this empty line.

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -19,18 +26,39 @@ class AccountPage extends React.Component<any, IAccountState>{
...
+                    this.state.usdBalance*data.rates.USD + 
+                    this.state.yenBalance*data.rates.JPY
+                })
+            }
+        })
+    }
+    
     componentDidMount(){
-
+
```
ed2298: Please remove this empty line.

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -171,10 +196,18 @@
 
 .transform-none{
     transform: none;
+    margin-left:30%;
+    position:absolute;
+    bottom:3%;
+
```
ed2298: Remove this empty line

Server/VikingVault.API/Controllers/AccountsController.cs
```diff
@@ -27,7 +27,10 @@ public AccountsController(IAccountService accService)
         [HttpGet]
         public ActionResult<UserAccount> Get()
         {
-            UserAccount userAccount = _accService.FindById();
+            var token = Request.Headers["x-access-token"];
+
```
ed2298, line 31: Remove this empty line. Also, you could  rename _accService into _accountService.

Server/VikingVault.DataAccess/VikingVaultDbContext.cs
```diff
@@ -15,6 +15,7 @@ public VikingVaultDbContext(DbContextOptions<VikingVaultDbContext> options)
         
         public DbSet<User> User { get; set; }
         public DbSet<BankAccount> BankAccount { get; set; }
+        public DbSet<Card> Card { get; set; }
```
ed2298: Please rename this variable into "Cards", since it is a set of cards.

Server/VikingVault.Services/AccountService.cs
```diff
@@ -2,26 +2,70 @@
...
-        public UserAccount FindById()
+        public UserAccount GetUserAccount(StringValues token)
         {
-            return null;
+            var tokenObject = new JwtSecurityToken(token);
+            string userId = tokenObject.Payload["Id"].ToString();
+
+            User returnedUser = _dbContext.User.SingleOrDefault(u => u.Id == int.Parse(userId));
+            Card card = _dbContext.Card.SingleOrDefault(u => u.UserId == int.Parse(userId));
+
```
ed2298: Remove this empty line

Server/VikingVault.Services/AccountService.cs
```diff
@@ -2,26 +2,70 @@
...
+                        break;
+                    case "Usd":
+                        userAccount.UsdBalance = ba.Balance;
+                        break;
+                    case "Yen":
+                        userAccount.YenBalance = ba.Balance;
+                        break;
+                }
+            }
+
```
ed2298: Remove this empty line, too.

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -1,37 +1,63 @@
...
 
 interface IAccountState{
-    balance: number
-    card: any
-    transactions: []
+    firstName: string
+    lastName: string
+    cardNumber: string
+    expirationDate: string
+    ronBalance: number
```
Slitthe: It would make more sense to group the balances in an object since they have multiple properties which are related to each other.

`interface IAccountBalance { eurBalance: number; usdBalance: number }.. blah blah`

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -1,37 +1,63 @@
...
         }
     }
-
+    
+    updateExchangingRates=() => {
+        fetch("https://api.exchangeratesapi.io/latest?base=RON").
+        then(response =>{
+            return response.json();
+        }).then((data:any)=>{
+            if(data!=null){
```
Slitthe: It's best to use `===` and `!==` in JavaScript as it's a bit more strict than `==` and `!=` which can cause errors

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -40,6 +66,7 @@ class AccountPage extends React.Component<any, IAccountState>{
         }
         else
         {
+            setInterval(this.updateExchangingRates, 5000);
```
Slitthe: You should extract this timeout period either in a constant in this file, or store it in the constants file, then when you want to change the timeout interval, you simply go there and change it rather than hunt it down to see where it's used

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -51,41 +78,101 @@ class AccountPage extends React.Component<any, IAccountState>{
                     if(response.status === 404){
                         this.setState(oldState => {
```
Slitthe: You don't need to use the `oldState` kind of calling the `setState` if you don't actually make use of the `oldState` argument. Usually, an unused variable/param is showed as grayed out in Visual Studio Code, this is usually a good sign that you should get rid of that.

In this case just call `setState({ key: value })` like this

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -51,41 +78,101 @@ class AccountPage extends React.Component<any, IAccountState>{
...
+                                ronBalance: userData.ronBalance,
+                                eurBalance: userData.eurBalance,
+                                usdBalance: userData.usdBalance,
+                                yenBalance: userData.yenBalance
+                            }
+                        )
+                    }
+                })
                 .catch(error => {
                     this.setState(oldState => {
```
Slitthe: Same here, you don't need to call `setState` with the `oldState` object if you don't use it

ClientApp/viking-vault/src/ConstantVariables.js
```diff
@@ -1,5 +1,8 @@
+import { timeout } from "q";
+
 const constants = {
-    baseUrl: "https://localhost:44323/api/"
+    baseUrl: "https://localhost:44323/api/",
+    timeoutInterval: 5000
```
vasi-axinte: What is this timeoutInterval for?

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -1,37 +1,74 @@
...
         this.state={
-            balance: 0,
-            card: {},
-            transactions: [],
-            isPresent: false,
+            firstName: "",
+            lastName: "",
+            cardNumber: "",
+            expirationDate: "",
+            accountsBalance: defaultBalances,
```
vasi-axinte: Why do we but defaultBalances in accountsBallance? These names aren`t really clear.

ClientApp/viking-vault/src/components/styles.css
```diff
@@ -50,36 +52,23 @@
...
+    min-width:400px;
+    width:400px;
+    height:250px;
     box-shadow: 1px 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(7, 7, 7, 0.19);
-    border-radius: 25px;
-    transition: 0.5s;
+    border-radius: 15px;
 }
 
 .info{
```
vasi-axinte: What kind of info? Please add some context.

Server/VikingVault.Services/AccountService.cs
```diff
@@ -2,26 +2,68 @@
...
         {
             _dbContext = dbContext;
-            accountsList = new List<UserAccount>();
         }
 
-        public UserAccount FindById()
+        public UserAccount GetUserAccount(StringValues token)
         {
-            return null;
+            var tokenObject = new JwtSecurityToken(token);
```
vasi-axinte: Method too long. Please refactor it.

vasi-axinte (changes requested):

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -49,43 +87,101 @@ class AccountPage extends React.Component<any, IAccountState>{
...
-                                isPresent: true
+                                isPresent: false
                             }
                         })
                     }
-                    response.json();})
-                .then(data=>{})
+                    return response.json();})
+                .then((userData:any)=>{
+                    console.log(userData)
```
crisfsabau: Please delete this console.log

ClientApp/viking-vault/src/components/AccountPage.tsx
```diff
@@ -49,43 +83,102 @@ class AccountPage extends React.Component<any, IAccountState>{
...
+
+    showUnavailableCard(){
+        return (
+        <div className="card-picture-container">
+            <img className="card-available" src={inexistentCard} alt=""></img>
+        </div>)
+    }
+
+    accountsInformation(){
+        return <div style={{width:'100%'}}>
```
crisfsabau: You shouldn't use inline style, use a class instead

Server/VikingVault.Services/AccountService.cs
```diff
@@ -2,26 +2,74 @@
...
         }
 
-        public UserAccount FindById()
+        public UserAccount GetUserAccount(StringValues token)
         {
-            return null;
+
+            int userId = FindUserIDByToken(token);
+            User returnedUser = _dbContext.User.SingleOrDefault(u => u.Id == userId);
+            Card card = _dbContext.Cards.SingleOrDefault(u => u.UserId == userId);
```
crisfsabau: What happens if returnedUser is null? you should retrieve card entity from db only if user is not null, otherwise u.UserId will throw an exception

moleguaco99: Returned user cannot ever be null, as this function is not called, if an user is not logged in

crisfsabau: What about if you test it from Postman, send a token for which the userId is not in db?

moleguaco99: It would just throw some exception from the token security, I shall catch it

vasi-axinte (approved):

PR #19, opened 2019-08-13
Fixed registration

robertpin:


Server/VikingVault.DataAccess/Migrations/20190813074007_DroppedProfilePages.cs
```diff
@@ -0,0 +1,73 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace VikingVault.DataAccess.Migrations
+{
+    public partial class DroppedProfilePages : Migration
```
vasi-axinte: Is this Migration related to the registration problem?

vasi-axinte (changes requested):

Server/VikingVault.DataAccess/Migrations/20190813074007_DroppedProfilePages.cs
robertpin: No. I removed it

vasi-axinte (approved):

PR #26, opened 2019-08-13
Fixed a bug in the exchange feature

OlahCatalina:


ClientApp/viking-vault/src/Router.tsx
```diff
@@ -78,7 +78,7 @@ class Router extends React.Component<any, IState> {
             <Route path="/register/" exact component={RegisterForm} />
             <Route path="/login" exact component={LoginForm} />
             <Route path="/" exact render={() => this.makeRedirect()}/>
-            <Route path="/user" component={UserPage} />
+            <Route path="/user" exact component={UserPage} />
             <Route path="/user/exchange" exact component={ExchangeForm}/>
             <Route path="/admin" render = { () => this.makeAdminRedirect()} />
         </BRouter>
```
vasi-axinte: What is a BRouter?

vasi-axinte (changes requested):

ClientApp/viking-vault/src/Router.tsx
OlahCatalina: It's a **BrowserRouter**, "obviously" :)))
Changed it.

robertpin (approved):

Server/VikingVault.Services/BankAccountService.cs
```diff
@@ -46,9 +46,9 @@ public List<BankAccount> GetBankAccounts(string token)
             return bankAccounts;
         }
 
-        public BankAccount ChangeBalance(string email, UpdateBankAccountModel updatedBankAccount)
+        public BankAccount ChangeBalance(UpdateBankAccountModel updatedBankAccount)
```
vasi-axinte, line 49: This method is named ChangeBalance but it only increments the value from the balance.

vasi-axinte (changes requested):

Server/VikingVault.Services/BankAccountService.cs
OlahCatalina: It could also decrement it.

ClientApp/viking-vault/src/ExchangeForm.tsx
```diff
@@ -203,19 +203,24 @@ class ExchangeForm extends React.Component<any, IExchangeFormState> {
...
+        else if(this.state.toExchangeAmount <= 0) {
             this.setState({
                 openModal: true,
                 modalMessage:  "The amount of money to exchange must be a positive number!"
             })
         }
+        else if(this.state.toExchangeAmount > this.state.availableAmountFromCurrency) {
+            this.setState({
+                openModal: true,
+                modalMessage:  "Not enough funds to execute the exchange!"
```
Slitthe: I think that perform should be used here instead of execute

ClientApp/viking-vault/src/ExchangeForm.tsx
```diff
@@ -233,15 +238,18 @@ class ExchangeForm extends React.Component<any, IExchangeFormState> {
                     body: JSON.stringify([  
                         {  
                             "CurrencyType": this.state.fromCurrency,
```
Slitthe: No need to make the properties of the JSON object PascalCase as ASP.NET automatically does that for you

ClientApp/viking-vault/src/ExchangeForm.tsx
OlahCatalina: 🔪 execute

OlahCatalina: Done, changed it, thank you.

ClientApp/viking-vault/src/ExchangeForm.tsx
OlahCatalina: sooo, should it be "currencyType"?

Slitthe: Yeah, pretty much, and when the data is mapped to C# objects, it will be accessible as CurrencyType

OlahCatalina: Changed it, thank you!

vasi-axinte (approved):

PR #27, opened 2019-08-13
Int2019 27 admin dashboard delete user

vladbuda:


ClientApp/viking-vault/src/display_user_list/UserData.tsx
```diff
@@ -24,59 +34,92 @@ const defaultUser = {
...
-                id: props.id,
-                firstName: props.firstName,
-                lastName: props.lastName,
-                address: props.address,
-                email: props.email,
-                pictureLink: props.pictureLink,
-                cardNumber: props.cardNumber,
-                expirationDate: props.expirationDate
-            };
+                user : {
```
robertpin, line 50: Please align object properly

ClientApp/viking-vault/src/display_user_list/AdminPage.tsx
```diff
@@ -84,6 +84,16 @@ class AdminPage extends React.Component<any, IProfileData>{
         }
     }
 
+    private deleteUserFromComponent = (email : string) =>{
```
vasi-axinte, line 87: shouldn`t this method remove a user from a list or smth? why are you filtering that list instead of deleting the user?

ClientApp/viking-vault/src/display_user_list/UserData.tsx
```diff
@@ -24,59 +34,96 @@ const defaultUser = {
...
+                        <p className = "expiration-date-on-card">{this.state.user.expirationDate}</p>
+                        <p className = "card-number-on-card">{this.state.user.cardNumber}</p>
                         <img src = {CardImg} className = "card-img"></img>
                    </div>        
 
         return null; 
     }
 
+    private handleDeleteUser = () =>{
+        this.setState((oldstate : any)=>({
```
vasi-axinte: why do you use any for oldState? (notice the naming)

vasi-axinte (changes requested):

ClientApp/viking-vault/src/display_user_list/UserData.tsx
vladbuda: That was the only way I managed to make this work, by specifying any for the type of oldState. Anyway, this was needed in the early version of this feature, but I removed it since we no longer need to toggle the state.

ClientApp/viking-vault/src/display_user_list/AdminPage.tsx
vladbuda: I think that using filter is a clean and elegant way of deleting a user from the list. There are many places where people advise using filter for deleting an entry from a list. Also, I didn't notice any explicit remove or delete functions. Another popular approach is to retrieve the index of the element and then split the list, but I think that using filter is neater.

ClientApp/viking-vault/src/display_user_list/UserData.tsx
```diff
@@ -24,59 +34,97 @@ const defaultUser = {
...
+                        <p className = "name-on-card">{this.state.user.firstName} {this.state.user.lastName}</p>
+                        <p className = "expiration-date-on-card">{this.state.user.expirationDate}</p>
+                        <p className = "card-number-on-card">{this.state.user.cardNumber}</p>
                         <img src = {CardImg} className = "card-img"></img>
                    </div>        
 
         return null; 
     }
 
+    private handleDeleteUser = () =>{
```
Slitthe: For simple methods you can write them in one line (don't overuse this though, but in this case it's okay)
 `private handleDeleteUser = () =>this.setState({ openDeleteUserModal: true });`

ClientApp/viking-vault/src/display_user_list/UserData.tsx
```diff
@@ -24,59 +34,97 @@ const defaultUser = {
...
+        this.setState({
+            openDeleteUserModal : true
+        }
+        );
+    }
+
+    private closeDeleteUserModal = () =>{
```
Slitthe: Same here

Server/VikingVault.Services/UserService.cs
```diff
@@ -63,5 +64,24 @@ private BankAccount CreateBankAccount(User user, String currencyType)
                 Balance = 0.0f
             };
         }
+
+        public void DeleteUser(UserEmail userEmail)
```
Slitthe, line 68: Why does this file has the same changes as the `int2019 28 admin dashboard add money` pull request, or the other way around?

vasi-axinte (approved):

PR #28, opened 2019-08-13
Int2019 26 admin dashboard attach card

ed2298:


ClientApp/viking-vault/src/AttachCardModal.css
```diff
@@ -0,0 +1,61 @@
+.modal-content.attach-card {
+    border-radius: 0.95rem;
+}
+
+.form-group.attach-card {
+    width: 86%;
+    margin-left: 7%;
```
Slitthe: Are you trying to center the element by using this margin?

ClientApp/viking-vault/src/AttachCardModal.css
```diff
@@ -0,0 +1,61 @@
...
+    padding: 0rem;
+    height: 53px;
+}
+
+.btn.attach-card {
+    width: 100%;
+    min-height: 100%;
+    margin-right: 0rem;
+    margin-left: 0rem;
+    border-radius: 0rem;
```
Slitthe: You don't need to specify unit size for `0` values.

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
+import React from "react";
+import { constants } from "./ConstantVariables";
+import './AttachCardModal.css';
+
+const baseUrl = constants.baseUrl;
+var regex = /^([+-]?[1-9]\d*|0)$/;
```
Slitthe: Please avoid using `var`s in JS as it's an older kind of declaring an variable, use `let` and `const` instead. (they behave differently and you should only use `var` if you want the functionality that the var provides or if you need to support an older browser and you don't transpile your code for some reason

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+interface IModalProps {
+    open: boolean;
+    modalClose: any;
```
Slitthe: I think that this would be better named `onModalClose` or `modalCloseHandler` as it works as an event

ClientApp/viking-vault/src/AttachCardModal.tsx
Slitthe: Also, please find a better name for this regex :), what do you try to match by this?

ClientApp/viking-vault/src/AttachCardModal.css
ed2298: Yes, that's what I wanted to do. I'll modify the seventh line, replacing it with margin: auto.

ClientApp/viking-vault/package.json
```diff
@@ -10,8 +10,10 @@
     "@types/react-router-dom": "^4.3.4",
     "bootstrap": "^4.3.1",
     "react": "^16.8.6",
+    "react-datepicker": "^2.8.0",
```
robertpin: Remove this

ClientApp/viking-vault/package.json
```diff
@@ -10,8 +10,10 @@
...
     "react-dom": "^16.8.6",
     "react-modal": "^3.9.1",
+    "react-month-year-picker": "^1.0.1",
```
robertpin: Remove this

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+}
+
+interface ICardProperties {
+    cardNumber: string;
+    expirationDate: string;
+    ccv: number;
+}
+
+interface IFormState {
+    [key: string]: string | number | boolean | ICardProperties  | null;
```
robertpin: Why is there a number? There isn't any property of type number in your state

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+            cardNumber: this.state.card.cardNumber,
+            expirationDate:  new Date(parseInt(this.state.card.expirationDate, 10), new Date().getMonth(), 1),
+            ccv: this.state.card.ccv,
+            userId: userId,
+        };
+    }
+
+    private sendDataAndShowResponse = async () => {
+        const card = this.getCard(this.props.userId);
+        console.log(card);
```
robertpin: Remove console.log

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+        .then(response => {
+            if(response.status === 409) {
+              this.setState({
+                errorLabel: "There is already a card with this card number!"
+              });
+              setTimeout(() => {
+                this.setState({
+                  errorLabel: ""
+                });
+              }, 5000);
```
robertpin: Isn't 5000 too much?

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+      
+            if(response.status === 500) {
+              this.setState({
+                errorLabel: "Error. Please try again later!"
+              });
+              setTimeout(() => {
+                this.setState({
+                  errorLabel: ""
+                });
+              }, 2500);
```
robertpin, line 152: You could probably change the one above to 2500 as well

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+                                <div className="form-group attach-card">
+                                    <label className="form-label attach-card">Card number</label>
+                                    <input type="text" value={this.state.card.cardNumber} onChange={(e) => this.handleChange(e.target.value, "cardNumber")} required className="form-control accent-color"></input>
+                                    {this.state.card.cardNumber !== "" && this.cardNumberHasSixteenDigits().message !== "Ok" ? <pre className={this.cardNumberHasSixteenDigits().class}>{this.cardNumberHasSixteenDigits().message}</pre> : null}
+                                </div>
+                                <div className="form-group attach-card">
+                                    <label className="form-label attach-card">Expiration date</label>
+                                    <select value={this.state.card.expirationDate} onChange={ (e) => this.handleChange(e.target.value, "expirationDate") } required className="form-control accent-color">
+                                        <option value=" "></option>
+                                        <option value="2020">2020</option>
```
robertpin: These values should not be hardcoded but created from the current year

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+                message: "Ok",
+                class: "alert alert-success"
+            }
+        return  {
+            message: "CCV must have 3 digits",
+            class: "alert alert-danger"
+        }
+    }
+
+    private expirationDateIsAfter2019 = () => {
```
robertpin: Should be renamed to exiprationDateIsAfterCurrentYear

Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
```diff
@@ -121,20 +121,6 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasName("Email");
 
                     b.ToTable("User");
-
```
robertpin: This should not be removed

robertpin (changes requested):

ClientApp/viking-vault/src/AttachCardModal.css
Slitthe: Yes, please use `margin: auto` when you want to center something instead of using percentages margins. Alternatively, you can also use flexbox to center both vertically and horizontally

ClientApp/viking-vault/src/AttachCardModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+            }
+        return  {
+            message: "Expiration Date must be after at least one year from the current date!",
+            class: "alert alert-danger"
+        }
+    }
+
+    private mandatoryFieldsCompletedCorrectly = () => {
+        let val = true;
+        if(this.cardNumberHasSixteenDigits().message !== "Ok") {
```
Slitthe: I'd replace all of this to something like this:


const validationFunctions = [this.cardNumberHasSixteenDigits, this.expirationDateIsAfterCurrentYear, this.ccvHasThreeDigits];

let val = false;

val = validationFunctions.all(functionItem => functionItem().message === "Ok")

return val;

In this way if you wanna add another validation function, you simply add it to the array (this actual code logic might be broken, but the idea should be that you store the validation functions in an array and then loop over them, call them and check for the `.message`

PR #29, opened 2019-08-13
Admin display users fixing bugs

radulambrino:


ClientApp/viking-vault/src/Router.tsx
```diff
@@ -60,17 +67,25 @@ class Router extends React.Component<any, IState> {
     {
         if(this.state.isAdmin === null)
         {
+            console.log("REQUEST FOR IS ADMIN SENT" + this.state.isAdmin);
```
OlahCatalina: Please remove this :)

ClientApp/viking-vault/src/display_user_list/DisplayUsers.css
```diff
@@ -67,29 +72,33 @@
     margin-top: 1%;
 }
 
-#user_name{
+#user-name{
```
crisfsabau: you shouldn't use ids for styling, id is using for identifying an html element. Use a class instead. Same for card number below :)

ClientApp/viking-vault/src/Router.tsx
```diff
@@ -63,14 +64,21 @@ class Router extends React.Component<any, IState> {
...
                 if(this.state.isAdmin)
                     return <AdminPage/>;
-                else return <Redirect to = "/login" />
-            },200);
+                else 
+                    return <Redirect to = "/login" />;
+            },500);
+        }
+        else
+        {
```
Slitthe: In the JS community, it's a sort of an unwritten rule/convention to write `{` on the same line

PR #30, opened 2019-08-14
Redirect from SideBar buttons

OlahCatalina:
Buttons from SideBar redirect to certain pages


Server/VikingVault.DataAccess/Migrations/20190814073151_SeedAdmin.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace VikingVault.DataAccess.Migrations
+{
+    public partial class SeedAdmin : Migration
```
vasi-axinte: This migration does nothing. Please remove it.

vasi-axinte (changes requested):

Server/VikingVault.DataAccess/Migrations/20190814073151_SeedAdmin.cs
OlahCatalina: Done 👍

vasi-axinte (approved):

PR #33, opened 2019-08-14
Int2019 improvement use picture link for icon

moleguaco99:


ClientApp/viking-vault/src/ProfilePicture.css
```diff
@@ -2,17 +2,16 @@
...
-    width: 20%;
-    height: 100%;
+    margin: 5%;
+    width: 150px;
+    height: 150px;
+    overflow: hidden;
+    border-radius: 50%;
 }
+
 #ProfilePicture
```
Slitthe: Please use the standard CSS ids and class naming which you use in your application, it's `kebab-case` I think?

ClientApp/viking-vault/src/ProfilePicture.css
```diff
@@ -2,17 +2,16 @@
 {
     position: relative;
     float: left;
```
Slitthe: It would be nice if you didn't use `float` as they are a bit nasty because of the height problem of the container when you use floats.

ClientApp/viking-vault/src/ProfilePicture.css
moleguaco99: Where were you when the first guy passed the pull request? :)))

ClientApp/viking-vault/src/ProfilePicture.tsx
```diff
@@ -1,14 +1,57 @@
...
-class ProfilePicture extends React.Component<any, any> 
+const API_URL = `${constants.baseUrl}userprofilepages`;
+
+interface IProfilePicture{
+    userProfilePicture: string
+}
+
+class ProfilePicture extends React.Component<any, IProfilePicture> 
 {
+    constructor(props: any){
```
Slitthe: You don't need a constructor if all you do is set up the state, simply use
` state = { userProfilePicture: profilePicture } `
in the class, same way as you'd define a class property

PR #37, opened 2019-08-20
ASAP:Int2019 fixing authorize user imports

moleguaco99:
I hope this is the right solution for import issues, I also added authorization for admin


Server/VikingVault.API/Controllers/AdminController.cs
```diff
@@ -34,6 +35,7 @@ public ActionResult<bool> IsAdmin()
 
         [Route("[action]")]
         [HttpGet]
+        [AuthorizeAdmin]
```
vasi-axinte: Instead of having a filter for  each role, it would be better to have 1 authorization filter that gets as input the role (ex: Authorize("admin")).

Server/VikingVault.API/SecurityFilters/AuthorizeAdmin.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
```
vasi-axinte: Please find a better solution than referencing AspNetCore.Mvc in your project.

vasi-axinte (changes requested):

Server/VikingVault.API/SecurityFilters/AuthorizeUser.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        void IAuthorizationFilter.OnAuthorization(AuthorizationFilterContext context)
+        {
+            try
+            {
+                string token = context.HttpContext.Request.Headers["x-access-token"].ToString();
+                var tokenObject = new JwtSecurityToken(token);
+                switch (Role)
+                {
+                    case "user":
```
vasi-axinte: This should be changes to use your role enum.

vasi-axinte (changes requested):

PR #39, opened 2019-08-20
Fixed admin dashboard delete user

vladbuda:


ClientApp/viking-vault/src/display_user_list/AdminPage.tsx
```diff
@@ -106,7 +106,7 @@ class AdminPage extends React.Component<any, IProfileData>{
                 <TopBar/>
                 <UserIcon></UserIcon>
```
Tedavid: If `<UserIcon>` component does not receive children as in: `<tag> children </tag>`, consider replacing with self closing tag `<UserIcon />`

ClientApp/viking-vault/src/display_user_list/DeleteUserModal.tsx
```diff
@@ -14,7 +14,7 @@ class DeleteUserModal extends React.Component<IDeleteUserModalProps, any>{
     }
 
     private closeModal = () =>{
```
Tedavid: If this method's only job is to call a parent method received in props, please remove it. 
A props received parent method can be called from any event (e.g. `onClick`).

ClientApp/viking-vault/src/display_user_list/DeleteUserModal.tsx
```diff
@@ -14,7 +14,7 @@ class DeleteUserModal extends React.Component<IDeleteUserModalProps, any>{
...
-        this.props.closeModal(false);
+        this.props.closeModal();
     }
 
     private handleDeleteUser = () =>{
```
Tedavid: Same ^

ClientApp/viking-vault/src/display_user_list/DeleteUserModal.tsx
```diff
@@ -4,8 +4,8 @@ import './DeleteUser.css';
 interface IDeleteUserModalProps{
     open: boolean;
     deletedUserName: string;
-    closeModal: any;
-    deleteUser: any;
+    closeModal: () => void;
+    deleteUser: () => void;
 }
 
 class DeleteUserModal extends React.Component<IDeleteUserModalProps, any>{
```
Tedavid: If there is no state in this component, do we really need a class-based component?
A functional component would be more appropriate.

Take a look over the following: 

- [React documentation regarding functional components](https://reactjs.org/docs/components-and-props.html)
- [What are they | Why are they used](https://programmingwithmosh.com/react/react-functional-components/)

PR #40, opened 2019-08-20
ASAP:Changed frontend folder structure

robertpin:


ClientApp/viking-vault/src/AdminDashboard/UserData.tsx
```diff
@@ -1,14 +1,13 @@
 import React from "react";
-import '../components/styles.css';
+import '../Resources/Components/styles.css';
```
vasi-axinte: Why is there a "Components" folder in Resources?

vasi-axinte (changes requested):

ClientApp/viking-vault/src/AdminDashboard/UserData.tsx
robertpin: For components that are used in other components. I could renamed it to "commons" if it would be more descriptive

vasi-axinte: That would be way better. Resources usually holds exactly that. Resources (i.e. images, json files and other stuff like that). You should have a "Common" folder only with components that are shared.

vasi-axinte (approved):

PR #41, opened 2019-08-20
ASAP: Int2019 59 update model for roles table

vladbuda:


Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
```diff
@@ -113,13 +147,15 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.Property<string>("PictureLink");
 
-                    b.Property<string>("Role");
+                    b.Property<int?>("RoleId");
```
vasi-axinte: Should this roleId be nullable?

vasi-axinte (changes requested):

Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
vladbuda: Yes, it should, but we decided to avoid sending the role from the frontend and pick the role from the backend accordingly. As a consequence, we needed to set this field on not required in order to be able to use this model hence the nullable property. In practice, this field will never be null because it will be automatically completed on the backend when a certain user is created -- regular user or company in this case.

vasi-axinte: If it will never be null, then it shouldn`t be nullable :).

Server/VikingVault.DataAccess/VikingVaultDbContext.cs
```diff
@@ -31,10 +33,31 @@ public void SeedAdmin(ModelBuilder modelBuilder)
...
+                    RoleId = 1
                 });
         }
 
+        public void SeedRoles(ModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<Role>().
+                HasData(new
+                {
+                    Id = 1,
```
Slitthe: It make more sense to assign int values to the enums (because that's what an enum basically is) and then use those as the Id values
`enum RoleEnum { Admin = 1}` `Id = (int)RoleEnum.Admin`

PR #42, opened 2019-08-20
Int2019 54 update model card

ed2298:


ClientApp/viking-vault/src/display_user_list/AttachCardModal.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { constants } from "./ConstantVariables";
+import { constants } from "../ConstantVariables";
 import './AttachCardModal.css';
 
 const baseUrl = constants.baseUrl;
```
vasi-axinte: This is not necessary. You could use constants.baseUrl as it is.

ClientApp/viking-vault/src/display_user_list/AttachCardModal.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { constants } from "./ConstantVariables";
+import { constants } from "../ConstantVariables";
```
vasi-axinte: Why is that file called "ConstantVariables"?

vasi-axinte (changes requested):

vasi-axinte (approved):

PR #43, opened 2019-08-21
ASAP: model for automatic payment

ed2298:


moleguaco99 (approved):

Server/VikingVault.DataAccess/Models/AutomaticPayment.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace VikingVault.DataAccess.Models
+{
+    public class AutomaticPayment
+    {
+        public int Id { get; set; }
+        [Required]
+        public User Company { get; set; }
```
Slitthe, line 12: Small thing, but it would make more sense to name this "RecievingCompany". Naming database columns descriptively is hugely important to understand how the application works.

Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
```diff
@@ -19,6 +19,31 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+
+                    b.Property<float>("Amount");
+
+                    b.Property<int>("CompanyId");
+
+                    b.Property<DateTime>("InitialPaymentDate");
+
+                    b.Property<DateTime>("LastPaymentDate");
+
+                    b.Property<int?>("PayingUserId");
```
Tedavid, line 36: Should property `PayingUserId` be nullable?

ed2298: If _PayingUserId_ would not be nullable, then it should have the _[Required]_ annotation above. But if both foreign keys: Company and PayingUser (which are of the same type (User)) have the  _[Required]_ annotation, the database will not be updated, and the following message appears : System.Data.SqlClient.SqlException: 'Introducing FOREIGN KEY constraint 'FK_AutomaticPayments_User_PayingUserId' on table 'AutomaticPayments' may cause cycles or multiple cascade paths. Specify ON DELETE NO ACTION or ON UPDATE NO ACTION, or modify other FOREIGN KEY constraints.
Could not create constraint or index. See previous errors.'

Therefore, I decided to remove the annotation on PayingUser, hence, this column is nullable.

PR #44, opened 2019-08-22
Int2019 46 add service company

robertpin:


Server/VikingVault.Services/CompanyService.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public CompanyService(VikingVaultDbContext dbContext, IBankAccountService bankAccountService)
+        {
+            _dbContext = dbContext;
+            _bankAccountService = bankAccountService;
+        }
+
+        public User CreateCompany(CompanyDTO company)
+        {
+            Role companyRole = _dbContext.Roles.SingleOrDefault(role => role.Type == RoleEnum.Company.ToString());
```
crisfsabau: use var

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -0,0 +1,71 @@
...
+            var presentUser = _dbContext.User.SingleOrDefault(user => user.FirstName.ToLower() == company.Name.ToLower());
+            if (presentUser != null)
+            {
+                throw new CompanyServiceException("Company already exists");
+            }
+            User companyUser = new User
```
crisfsabau: use var

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            try
+            {
+                _dbContext.Add(companyUser);
+                _bankAccountService.CreateBankAccount(this.CreateBankAccount(companyUser, "Ron"));
+
+                _dbContext.SaveChanges();
+            } catch (Exception e)
+            {
+                if (e is DbUpdateException || e is DbUpdateConcurrencyException || e is BankAccountServiceException)
```
crisfsabau: if it's not any of this exceptions what happens? the exception is swallowed and it is hard to figure what went wrong. you should throw regarding the exception type

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -0,0 +1,71 @@
...
+                {
+                    throw new CompanyServiceException("Internal server error");
+                }
+            }
+
+            return companyUser;
+        }
+
+        private BankAccount CreateBankAccount(User user, String currencyType)
```
crisfsabau: use string instead of String

Server/VikingVault.API/Controllers/CompanyController.cs
```diff
@@ -0,0 +1,44 @@
...
+        [HttpPost]
+        public ActionResult<User> Post([FromBody] CompanyDTO company)
+        {
+            try
+            {
+                return Ok(_companyService.CreateCompany(company));
+            }
+            catch(CompanyServiceException e)
+            {
+                if(e.Message == "Company already exists")
```
Slitthe: It's better to create a more specific exception here and catch that instead of checking for the message. What if you want to change the message of that exception?

PR #45, opened 2019-08-22
ASAP:View all companies done

robertpin:


Server/VikingVault.Services/CompanyService.cs
```diff
@@ -64,5 +64,33 @@ private BankAccount CreateBankAccount(User user, string currencyType)
                 Balance = 0.0f
             };
         }
+
+        public List<CompanyDataDTO> GetAllCompanies()
+        {
+            List<CompanyDataDTO> companiesData = new List<CompanyDataDTO>();
```
crisfsabau: please use var

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -64,5 +64,33 @@ private BankAccount CreateBankAccount(User user, string currencyType)
...
+            try
+            {
+                List<User> companies = _dbContext.User.Where(user => user.Role.Type == RoleEnum.Company.ToString()).ToList();
```
crisfsabau: please use var

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -64,5 +64,33 @@ private BankAccount CreateBankAccount(User user, string currencyType)
...
+                foreach (User company in companies)
```
crisfsabau: please use var

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -64,5 +64,33 @@ private BankAccount CreateBankAccount(User user, string currencyType)
...
+                {
+                    BankAccount account = _dbContext.BankAccount
```
crisfsabau: please use var

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -64,5 +64,33 @@ private BankAccount CreateBankAccount(User user, string currencyType)
...
+                        Id = company.Id,
+                        Name = company.FirstName,
+                        Address = company.Address,
+                        Balance = account.Balance
+                    });
+                }
+            }
+            catch (Exception e)
+            {
+                throw new CompanyServiceException();
```
crisfsabau: Maybe we should think if we want to return the message from the exception

PR #46, opened 2019-08-22
Int2019 29 transaction summary in pdf

OlahCatalina:
Transaction list downloadable as PDF


ClientApp/viking-vault/src/UserDashboard/DownloadTypeForm.css
```diff
@@ -0,0 +1,40 @@
+.download-icon {
+    height: 7%;
+    width: 7%;
+    position: relative;
+    border: none;
+}
+
+.download-button:focus,.download-button:active {
+    outline: none !important;
```
Slitthe: `!important` is like dropping an atomic bomb 💣☢to kill a rat 🐀, it's really really best to avoid it. See a bit about CSS selector specificity to see how you could avoid this

ClientApp/viking-vault/src/UserDashboard/DownloadTypeForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                })
+                .then((response) => response.blob())
+                .then((blob) => {
+                    if(blob != null) {
+                    const url = window.URL.createObjectURL(new Blob([blob]));
+                    const link = document.createElement('a');
+                    link.href = url;
+                    link.setAttribute('download', `Viking Vault Report.pdf`);
+                    document.body.appendChild(link);
```
Slitthe: Hmm... why do you try to manipulate the DOM directly instead of doing it through the React library?

OlahCatalina: I deleted this line
**document.body.appendChild(link);**

Copy-paste... Hihi... Thanks!

ClientApp/viking-vault/src/UserDashboard/DownloadTypeForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    link.click(); }
+                })
+            }
+        }
+       
+
+
+    handlePDFRequestPerDay = () => {
```
Slitthe: It seems to me like these four `handlePDFRequestPer{something}` could be made into a single function-creating factory.
Like this 

const multiplyBy = function(byHowMuch) {
	return function(amount) {
		return amount * byHowMuch;
	}
}

const multiplyByThree = multiplyBy(3);
multiplyByThree(5);

ClientApp/viking-vault/src/UserDashboard/DownloadTypeForm.css
OlahCatalina: I deleted those and it seems to work just the same... Thank you!

ClientApp/viking-vault/src/UserDashboard/DownloadTypeForm.tsx
OlahCatalina: Yep, forgot. Done, thanks!

moleguaco99 (approved):

Server/VikingVault.Services/PDFGeneratorService.cs
```diff
@@ -0,0 +1,112 @@
...
+        private ITransactionService _transactionService;
+
+        public PDFGeneratorService(ITransactionService transactionService)
+        {
+            _transactionService = transactionService;
+        }
+
+        public string GetTimeFilterHeaderText(string timeFilter)
+        {
+            if (timeFilter == "day")
```
Slitthe, line 28: Please make the code blocks consistent to the .NET styling (each block on their own line)
{
}

Server/VikingVault.Services/PDFGeneratorService.cs
```diff
@@ -0,0 +1,112 @@
...
+            PdfBitmap wirtekImage = new PdfBitmap(File.OpenRead("Resources\\logo_wirtek.png"));
+            graphics.DrawImage(wirtekImage, -35, 0, 300, 45);
+            graphics.DrawString("Report", fontReportText, brush, new PointF(32, 45));
+            graphics.DrawString(timeFilterHeaderText, fontTimeText, PdfBrushes.Orange, new PointF(50, 80));
+            graphics.DrawString("Today's date: " + DateTime.Today.Date.ToShortDateString(), fontDateText, brushDate, new PointF(380, 110));
+        }
+
+        public List<Transaction> FilterTransactionsByTime(List<Transaction> transactionList, string timeFilter)
+        {
+            if (timeFilter == "day") {
```
Slitthe: Samesies as above

ClientApp/viking-vault/src/UserDashboard/DownloadTypeForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+                })
+            }
+        }
+
+    handlePDFRequestPer = (timeFilter: string) => {
+        this.setState({
+            timeFilter: timeFilter,
+            isOpen: !this.state.isOpen
+        }, this.generatePDF);
+        return null;
```
Slitthe: Kind of redundant to return `null` here, if a method is void, just don't return return (and don't look back 😅), eventually use the typescript type of `void` as the function's return type to explicitly tell that this doesn't return any data

OlahCatalina: I removed it, thankies.

Server/VikingVault.Services/PDFGeneratorService.cs
OlahCatalina: Changed it. Thank you!

Server/VikingVault.Services/PDFGeneratorService.cs
OlahCatalina: Yup, done.

PR #48, opened 2019-08-22
Finished delete company

robertpin:


ClientApp/viking-vault/src/AdminDashboard/CompanyComponents/DeleteCompany.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import trash from "../../Resources/images/delete.png";
```
Slitthe: Stop importing trash, maybe rename it to `deleteIcon`?

PR #49, opened 2019-08-22
ASAP:Fixed exchange

robertpin:


ClientApp/viking-vault/src/UserDashboard/ExchangeForm.tsx
```diff
@@ -228,8 +249,8 @@ class ExchangeForm extends React.Component<any, IExchangeFormState> {
         }
         else
         {
-            var amount = this.state.toExchangeAmount;
-            var otherParty = balance.toFixed(3) + " " + this.state.toCurrency;
+            var amount = this.state.exchangedAmount.toFixed(2);
+            let otherParty = `Exchanged ${this.state.toExchangeAmount} ${this.state.fromCurrency}`;
```
vasi-axinte: What does otherParty stands for?

ClientApp/viking-vault/src/UserDashboard/ExchangeForm.tsx
```diff
@@ -362,23 +385,31 @@ class ExchangeForm extends React.Component<any, IExchangeFormState> {
...
-                                        </tr>
-                                        <tr className="padding-design">
-                                            <td className="currency-decoration padding-design">{this.state.fromCurrency === "USD"? null : `USD`}</td>
-                                            <td className="table-design padding-design">{this.state.fromCurrency === "USD"? null : `${this.state.currencyToUsd}`}</td>
-                                        </tr>
-                                        <tr className="padding-design">
-                                            <td className="currency-decoration padding-design">{this.state.fromCurrency === "YEN"? null : `YEN`}</td>
-                                            <td className="table-design padding-design">{this.state.fromCurrency === "YEN"? null : `${this.state.currencyToYen}`}</td>
                                         </tr>
+                                        {this.state.fromCurrency === "RON" ? null :
```
vasi-axinte: It would be better to have an enum for currencies.

vasi-axinte (changes requested):

ClientApp/viking-vault/src/UserDashboard/ExchangeForm.tsx
robertpin: There are two parties involved in a transaction: current user and the other party

vasi-axinte: Why not sender and receiver then? It would make more sens.

robertpin: Because it can be: sender, receiver, company (that receives from automatic payments). So I don't this that Sender/Receiver/Company is a good column name

ClientApp/viking-vault/src/UserDashboard/ExchangeForm.css
```diff
@@ -88,15 +85,14 @@
     color: rgb(29, 84, 95);
     font-size: 120%;
     font-family: Helvetica;
-    font-weight: 750;
 }
 
 .white-box-background {
```
Slitthe: this `.white-box-background` is more than a white box with a background though, you also set the width and margin etc.. think of a better way to name CSS classes or split them into 2 classes, one for a specific kind of box styling (lets say one with a with background with a box shadow) and another one where you add things such as `margin` and all that positioning to be a more specific class, and then this becomes `.classOne.classTwo`

ClientApp/viking-vault/src/UserDashboard/ExchangeForm.css
```diff
@@ -88,15 +85,14 @@
...
     background-color: white;
     margin-top: 2%;
     margin-left: 13%;
```
Slitthe: Look up `margin: auto` :)

ClientApp/viking-vault/src/UserDashboard/ExchangeForm.tsx
vasi-axinte: Receiver works for both a company and a user.

vasi-axinte: Same goes with sender.

robertpin: So I should name the column in the table Sender/Receiver?

vasi-axinte: Yes, please.

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -4,15 +4,18 @@ import "./TransactionList.css";
...
 const baseUrl = constants.baseUrl;
 
 interface ITransaction {
+    user: IUserData;
     type: string;
     date: Date;
     amount: number;
     currency: string;
-    otherParty: string;
+    senderOrReceiver: IUserData;
```
vasi-axinte: Ever heard of SRP? :)

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -47,16 +50,16 @@ class TransactionList extends React.Component<any, IState> {
     }
 
     private formatOtherPartyString(transaction: ITransaction) {
-        if(transaction.type === "transfer") {
+        if(transaction.type.toLowerCase() === "transfer") {
```
vasi-axinte: Why didn`t you also create an enum for transactionType? As a rule of tumb, we don`t just put hardcoded strings through code.

Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
```diff
@@ -137,14 +137,18 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.Property<DateTime>("Date");
 
-                    b.Property<string>("OtherParty");
+                    b.Property<string>("Details");
+
+                    b.Property<int?>("SenderOrReceiverId");
```
vasi-axinte: This basically means that a transaction can exist without sender/receiver? Doesn`t sound good to me.

Server/VikingVault.Services/ScopedProcessingService.cs
```diff
@@ -68,8 +68,9 @@ private void AddPayingUserAutomaticTransaction(AutomaticPayment automaticPayment
                 Type = "Payment",
                 Date = currentDateTime,
                 Currency = "Ron",
-                Amount = automaticPayment.Amount,
-                OtherParty = automaticPayment.ReceivingCompany.FirstName
+                Amount = -automaticPayment.Amount,
```
vasi-axinte: The name of this service is confusing. Why Scoped, Why Processing? Also, this - added here and there confuse me since it is hard for me to understand why we`re doing it. If you leave the project and I`m going to continue it, things are going to be hard for me to understand.

vasi-axinte (changes requested):

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
robertpin: No 😒

robertpin: > Ever heard of SRP? :)

SenderReceiverParty? :))

Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
robertpin: If I make it required, this happens: Introducing FOREIGN KEY constraint 'FK_Transactions_User_SenderOrReceiverId' on table 'Transactions' may cause cycles or multiple cascade paths. Specify ON DELETE NO ACTION or ON UPDATE NO ACTION, or modify other FOREIGN KEY constraints.

Server/VikingVault.Services/ScopedProcessingService.cs
robertpin: 1. ScopedProcessing because that is what it looks like on Microsoft docs (so I've been told)
2. When we create the transaction we add a negative amount because the user is basically paying which mean that his balance changed with -amount

vasi-axinte: 1. Still, it`s confusing, since I cannot see the link between what it does and the name.
2. That should be formalized somehow, because it is prone to errors.

Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
vasi-axinte: You don`t need to have on cascade delete on all your foreign keys. Think about how can you write everything so that it doesn`t give you that error anymore.

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
vasi-axinte: senderOrReceiver has 2 purposes and it is not a good practice. Please rewrite your code or rename it such that it only has one purpose to exist.

Server/VikingVault.DataAccess/Migrations/VikingVaultDbContextModelSnapshot.cs
robertpin: Already tried. Can't make it work. Would appreciate some help

robertpin: Done

ClientApp/viking-vault/src/UserDashboard/TransactionList.css
```diff
@@ -10,6 +10,13 @@
...
 }
 
+.transactions table td.details {
+    max-width: 150px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+}
+
 .transactions h5 {
```
Slitthe: It's best to avoid using such specific html elements in your CSS selectors such as `.h5`, it would be better to replace this with a class

Server/VikingVault.Services/TransactionService.cs
```diff
@@ -47,41 +47,50 @@ public void AddTransactionsForTransferFunds(TransferFundsModel transferData)
...
         {
             int uid = int.Parse(userId);
-            return _dbContext.Transactions
-                .Where(t => t.User.Id == uid)
+            var transactions =  _dbContext.Transactions
+                .Include(t => t.Sender)
+                .Include(t => t.Receiver)
+                .Where(t => t.Sender.Id == uid || t.Receiver.Id == uid)
                 .OrderByDescending(t => t.Date)
                 .Take(5)
```
Slitthe: This `.Take()` amount should either be passed as an argument to the function,or extracted into a constant or you make two overloaded methods, one which gets the with the default amount, one where you pass in the amount.

Server/VikingVault.Services/ExecutePaymentService.cs
```diff
@@ -64,12 +64,14 @@ private void AddPayingUserAutomaticTransaction(AutomaticPayment automaticPayment
...
             Transaction transaction = new Transaction
             {
-                User = automaticPayment.PayingUser,
+                Sender = automaticPayment.PayingUser,
                 Type = "Payment",
                 Date = currentDateTime,
                 Currency = "Ron",
                 Amount = automaticPayment.Amount,
-                OtherParty = automaticPayment.ReceivingCompany.FirstName
+                //Sender = automaticPayment.PayingUser,
```
Slitthe: Delete this comment if you don't need it

Server/VikingVault.Services/ExchangeService.cs
```diff
@@ -38,12 +38,13 @@ public List<BankAccount> Exchange(string token, UpdateBankAccountModel bankAccou
...
-                User = returnedUser,
+                Sender = returnedUser,
                 Type = "Exchange",
-                Currency = bankAccountSell.CurrencyType,
+                Currency = bankAccountBuy.CurrencyType,
                 Date = DateTime.Now,
                 Amount = exchangeInfo.Amount,
-                OtherParty = exchangeInfo.CurrencyType
+                Receiver = returnedUser,
+                Details = exchangeInfo.CurrencyType
```
Slitthe, line 47: Why is this called `Details` when it stores the an `CurrencyType` ?

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -75,22 +81,40 @@ class TransactionList extends React.Component<any, IState> {
...
+    }
+
+    private formatAmount = (transation: ITransaction) => {
+        if(transation.type === "payment") {
+            return -transation.amount;
+        }
+        if(transation.amount > 0) {
+            return `+${transation.amount}`;
+        }
+        return transation.amount;
```
Slitthe: Seeing a lot of return statements is confusing, can you get rid of them and only use them once?

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -75,22 +81,40 @@ class TransactionList extends React.Component<any, IState> {
...
-        if(transactionType.toLowerCase() === "exchange") {
+        if(transactionType.toLowerCase() === transactionTypeEnum.exchange) {
             return exchange;
         }
         return shoppingCart;
     }
 
+    private formatDetails(tranasction: ITransaction) {
+        if(tranasction.type.toLowerCase() === transactionTypeEnum.exchange) {
+            return "Exchange";
```
Slitthe: Same as below

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -49,16 +53,18 @@ class TransactionList extends React.Component<any, IState> {
...
             }
-            return `To ${transaction.otherParty}`;
         }
-        if(transaction.type === "exchange") {
-            return `Exchanged ${transaction.otherParty.toUpperCase()}`;
+        if(transaction.type.toLowerCase() === transactionTypeEnum.exchange) {
+            return `Exchanged ${transaction.details.toUpperCase()}`;
         }
-        return transaction.otherParty;
+        return transaction.receiver.firstName;
```
Slitthe: Same as below

Server/VikingVault.Services/ExchangeService.cs
robertpin: Because exchangeInfo.currencyType actually stores details but it uses the same model as bankAccountSell

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
robertpin: There are like 3 of them but fine :)). Done

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -1,20 +1,24 @@
...
 import '../UserDashboard/DownloadTypeForm.css'
 import shoppingCart from '../Resources/images/shoppingcart.png';
 import userIcon from '../Resources/images/profileWhite.png';
 import exchange from "../Resources/images/money-exchange.png";
-import DownloadDropdown from './DownloadTypeForm'
+import { IUserData } from './../AdminDashboard/UserData';
+import DownloadDropdown from './DownloadTypeForm';
 
-const baseUrl = constants.baseUrl;
+const url = constants.baseUrl+"transaction";
```
crisfsabau: please rename it to transactionUrl

PR #50, opened 2019-08-23
INT2019 61 Block Services

robertpin:


ClientApp/viking-vault/src/UserDashboard/BlockedServices.tsx
```diff
@@ -0,0 +1,108 @@
...
+        this.checkUserHasCard();
+        this.checkCardIsBlocked();
+    }
+
+    checkUserHasCard = () => {
+        const token = sessionStorage.getItem("Authentication-Token");
+        if(token === null) {
+            return;
+        }
+        fetch(constants.baseUrl+"card", {
```
crisfsabau: You could extract constants.baseUrl+"card" in a constant, and name it accordingly

ClientApp/viking-vault/src/UserDashboard/BlockedServices.tsx
```diff
@@ -0,0 +1,108 @@
...
+            });
+        });
+    }
+
+    checkCardIsBlocked = () => {
+        const token = sessionStorage.getItem("Authentication-Token");
+        if(token === null) {
+            return;
+        }
+        fetch(constants.baseUrl+"blockedcard", {
```
crisfsabau: same as above

ClientApp/viking-vault/src/UserDashboard/BlockedServices.tsx
```diff
@@ -0,0 +1,108 @@
...
+        }).then(response => {
+            if(response.status !== 200) {
+                return null;
+            }
+            return response.json();
+        }).then(result => {
+            if(result === null) {
+                return;
+            }
+            console.log(result);
```
crisfsabau: remove console.log

ClientApp/viking-vault/src/UserDashboard/BlockedServices.tsx
```diff
@@ -0,0 +1,108 @@
...
+            this.setState({
+                isCardBlocked: result
+            });
+        });
+    }
+
+    returnBlockServiceView = () => {
+        let message = "";
+        if(this.state.userHasCard === false) {
```
crisfsabau, line 86: you could use if(!this.state.userHasCard)

ClientApp/viking-vault/src/UserDashboard/BlockedServices.tsx
```diff
@@ -0,0 +1,108 @@
...
+        return <React.Fragment>
+            <SideBar userType="user"/>
+            <TopBar/>
+            <UserIcon/>
+            <div className="white-container w-75 mr-auto ml-auto bg-white">
+                <h3 className="text-center m-5">{message}</h3>
+            </div>
+        </React.Fragment>
+    }
+
```
crisfsabau, line 100: remove empty line, leave just one

ClientApp/viking-vault/src/UserDashboard/BlockedServices.tsx
```diff
@@ -0,0 +1,108 @@
...
+    
+    render() {
+        return <div className="dark-background">
+            {this.state.userHasCard && !this.state.isCardBlocked? this.props.children : this.returnBlockServiceView()}
+        </div>;
+    }
+
```
crisfsabau: remove empty line

Server/VikingVault.Services/CardService.cs
```diff
@@ -46,5 +47,39 @@ public Card AttachCard(Card cardToAttach)
                 throw new DatabaseException("Database Error!");
             }
         }
+
+        public Card CheckUserHasCard(string token)
+        {
+            try
+            {
+                var tokenObject = new JwtSecurityToken(token);
+                string userId = tokenObject.Payload["Id"].ToString();
```
crisfsabau: use var

Server/VikingVault.Services/CardService.cs
```diff
@@ -46,5 +47,39 @@ public Card AttachCard(Card cardToAttach)
...
+                throw new CardServiceException(e.Message);
+            }
+            
+        }
+
+        public bool CheckCardIsBlocked(string token)
+        {
+            try
+            {
+                Card card = CheckUserHasCard(token);
```
crisfsabau: use var

ClientApp/viking-vault/src/AdminDashboard/AttachCardModal.tsx
```diff
@@ -120,12 +120,17 @@ class AttachCardForm extends React.Component<any, IFormState> {
     }
 
     private sendDataAndShowResponse = async () => {
+        const token = sessionStorage.getItem("Authentication-Token");
+        if(token === null) {
+            return;
+        }
         const card = this.getCard(this.props.userId);
         fetch(constants.baseUrl+"attach", {
```
crisfsabau: You could extract constants.baseUrl+"attach" in a constant, and name it accordingly

ClientApp/viking-vault/src/Common/styles.css
```diff
@@ -364,4 +364,21 @@
 
 .redirect-symbols:hover {
     text-decoration: none;
+}
+
+.white-container {
```
Slitthe: It's a bit weird for a class to be named as white container in the first place. It's better to name it according to what the element that the class is attached to represents. What if for example you want to change the color theme of the website, then you'd also have to change the class name.

But in this case it's also weird because it doesn't contain any white color anywhere in the class properties

ClientApp/viking-vault/src/UserDashboard/BlockedServices.tsx
robertpin: Not exactly, because userHasCard is initializes with null. Until we get the response from the backend we don't assume anything. Considering null is a falsy value !this.state.userHasCard would evaluate to true and the you-don't-have-a-card message would be displayed until the answer from the server is received

PR #51, opened 2019-08-23
Int2019 33 transfer funds

radulambrino:
Transfer Funds + RequestTransfer(sender view)


ClientApp/viking-vault/src/Common/Toggle.css
```diff
@@ -0,0 +1,68 @@
+.switch {
```
vasi-axinte: This is not suggestive enough and it should be changed to something more meaningful.

vasi-axinte (changes requested):

ClientApp/viking-vault/src/TransferFunds/TransferFundsPage.tsx
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    private setCurrency = (e: any) => {
+      this.setState({
+        currency: e.target.value
+      });
+      this.getTotalBalance();
+    }
+
+    handleChangerequestTransferState = () =>
```
vasi-axinte: ChangeRequest are two different words, so they should start with capital letter.

ClientApp/viking-vault/src/TransferFunds/TransferFundsPage.tsx
```diff
@@ -0,0 +1,328 @@
...
+                method: "GET",
+                headers: {
+                  'Accept': 'application/json',
+                  'Content-Type': 'application/json',
+                  'x-access-token': token.toString()
+                }
+            })
+            .then(response => response.json())
+            .then(result => {
+                if(this.state.currency === "RON") {
```
vasi-axinte: This method became too long. Should be refactored to be smaller. One can notice that all those if`s differ slightly, therefore they can be easily extracted to a method.

ClientApp/viking-vault/src/TransferFunds/TransferFundsPage.tsx
```diff
@@ -0,0 +1,328 @@
...
+                        totalBalance: result[3].balance.toFixed(2)
+                    });
+
+                }
+            });
+        }
+    }
+
+    transferMoney = () => {
+        var data = {
```
vasi-axinte: Same here. Method too long. In order to understand it, I need to read a lot of code. That`s a lot of effort.

ClientApp/viking-vault/src/TransferFunds/TransferFundsPage.tsx
```diff
@@ -0,0 +1,328 @@
...
+                  }
+                  this.handleIsButtonDisabled(); 
+              });
+          }
+        }
+    }
+
+    requestTransfer = () =>
+    {
+      var data = {
```
vasi-axinte: Same here.

Server/VikingVault.API/Controllers/TransferRequestsController.cs
```diff
@@ -0,0 +1,68 @@
...
+                        return Ok("Succesfully requested " + transferRequestData.Amount + "!");
+                    }
+                    else
+                    {
+                        return Ok("Select a different user to request money from!");
+                    }                   
+                }
+                else
+                {
+                    return NotFound("Request to server unsuccesful.");
```
vasi-axinte: The request was ok, the problem would be that it didn`t find something that was requested.

Server/VikingVault.Services/Abstractions/IBankAccountService.cs
```diff
@@ -11,5 +11,7 @@ public interface IBankAccountService
         List<BankAccount> GetBankAccounts(string token);
         void UpdateBankAccount();
         BankAccount ChangeBalance(UpdateBankAccountModel updatedBankAccount);
+        BankAccount AddMoneyToAccount(User user, string currency, float amountToBeAdded);
```
vasi-axinte: How comes that you named a method "AddMoneyToAccount" but no account was sent as parameter?

Server/VikingVault.Services/TransferFundsService.cs
```diff
@@ -0,0 +1,65 @@
...
+            _dbContext = dbContext;
+            _bankAccountService = bankAccountService;
+            _transactionService = transactionService;
+            _userCardService = userCardService;
+            _userService = userService;
+        }
+
+        public bool? TransferFunds(TransferFundsModel transferData)
+        {
+            int idSender = transferData.Sender.Id;
```
vasi-axinte, line 34: You have so many if`s in this method... You shall refactor it in order to use fewer if`s or at least, extract some of them to separate methods.

vasi-axinte (changes requested):

Server/VikingVault.API/Controllers/TransferRequestsController.cs
radulambrino: Fair enough. I was unsure about this kind of situations and should've asked you when you was here.

Server/VikingVault.Services/Abstractions/IBankAccountService.cs
radulambrino: In this case, I am using the method in the controller, so I wanted to pass the responsibility of creating the account model to the method. I thought of adding only one BankAccount parameter to the method, but in the end I chose to do it this way beacause it reduces the number of lines in the top level code. In this case do you suggest changing the method name or the signature?

ClientApp/viking-vault/src/Common/Toggle.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from "react";
+import './Toggle.css';
+import { constants } from "../Resources/Constants";
+
+interface IToggleState{
+    toggleSwitch: any
+}
+
+function Toggle (props: IToggleState)
+{
```
Slitthe: JavaScript convention is that the first code block is on the same line
blah {
}

ClientApp/viking-vault/src/TransferFunds/TransferFundsPage.tsx
```diff
@@ -0,0 +1,328 @@
...
+
+    handleChangedCardNumber = (e : any) =>
+    {
+        this.setState({
+            cardNumber: e.target.value
+        })
+    }
+
+    handleChangedTransferDetails = (e : any) =>
+    {
```
Slitthe: Same here and above and below (code block positioning in JavaScript)

ClientApp/viking-vault/src/TransferFunds/TransferFundsPage.tsx
```diff
@@ -0,0 +1,328 @@
...
+                    this.setState({
+                        totalBalance: result[1].balance.toFixed(2)
+                    });
+                }
+                if(this.state.currency === "USD") {
+                    this.setState({
+                        totalBalance: result[2].balance.toFixed(2)
+                    });
+                }
+                if(this.state.currency === "YEN") {
```
Slitthe: Please find a way to get rid of these multiple if's, they're really a bad practice as if they are hard to modify and it's also a lot of repeating code.

Proposed solution
-- create a dictionary for currencies and their indexes (or indices, whatever it's spelled) in the result array
-- return something like this `result[currenyMap[this.state.currency]].balance.toFixed(2)`
(of course also handle the default value, you can use Object.HasOwnProperty("propName") too see if an object has a property with that name)

Server/VikingVault.API/Controllers/TransferFundsController.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    User sender = _userService.GetUserFromToken(token);
+                    TransferFundsModel transferFundsData = transferFundsDTO.convertDTOtoTransferFundsModel(sender);
+
+                    _transferFundsService.TransferFunds(transferFundsData);
+                    return Ok("Succesfully transfered " + transferFundsData.AmountSent + "!");
+                }
+                else
+                {
+                    return NotFound("Request to server unsuccesful.");
```
Slitthe: Could you be more explicit with this error message?

Server/VikingVault.API/Controllers/TransferRequestsController.cs
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+            catch (Exception e)
+            {
+                if (e is NoCardAttachedToUserException || e is TransactionException || e is DatabaseException)
+                {
+                    return NotFound(e.Message);
+                }
+
+                return NotFound("Unknown error.");
```
Slitthe, line 64: Usually if something goes wrong but you don't know what's wrong it's an 500 Internal Server Error, a NotFound has the status code of 404 which simply menans that the resource that you've asked couldn't be found, but everything else worked as expected. But in here the NotFound is thrown at the end of the `catch( Exception)`

Server/VikingVault.Services/BankAccountService.cs
```diff
@@ -55,6 +55,35 @@ public BankAccount ChangeBalance(UpdateBankAccountModel updatedBankAccount)
...
+            var updatedAccount = new UpdateBankAccountModel
+            {
+                CurrencyType = currency,
+                Amount = -amountSubstracted,
+                Email = user.Email
+            };
+
+            BankAccount currentAccount = _dbContext.BankAccount.SingleOrDefault(account => account.CurrencyType == updatedAccount.CurrencyType && account.User == user);
+
+            if ( (currentAccount.Balance + updatedAccount.Amount) < 0 )
```
Slitthe: Wrap the if in some curly brackets please, consistency is good

Server/VikingVault.Services/UserCardService.cs
```diff
@@ -0,0 +1,59 @@
...
+                throw new DatabaseException("Server error.");
+            }
+        }
+
+        public int? FindUserIdByCardNumber(string cardNumber)
+        {
+            try
+            {
+                Card card = _dbContext.Cards.SingleOrDefault(c => c.CardNumber == cardNumber);
+                if (card == null)
```
Slitthe, line 38: Same here

Server/VikingVault.API/Controllers/TransferRequestsController.cs
radulambrino: Fair enough. Changed it

PR #54, opened 2019-08-23
Int2019 63 create notification model

robertpin:


Server/VikingVault.DataAccess/Models/Notification.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace VikingVault.DataAccess.Models
+{
+    public class Notification
+    {
+        [Required]
+        public User user { get; set; }
```
vasi-axinte: Variable naming doesn`t follow the c# naming standards.

Server/VikingVault.DataAccess/Models/Notification.cs
```diff
@@ -0,0 +1,19 @@
...
+        [Required]
+        public string Text { get; set; }
+        [Required]
+        public bool Read { get; set; }
```
vasi-axinte, line 17: Why do we have a boolean Read and Seen? Can`t we use only one?

vasi-axinte (changes requested):

Server/VikingVault.DataAccess/Models/Notification.cs
robertpin: Because it's read only if you check it as read, but when the user logs in we display something only if he has unseen notifications

robertpin: Done

vasi-axinte (approved):

PR #56, opened 2019-08-26
Int2019 49 create automatic payment

ed2298:


ClientApp/viking-vault/src/UserDashboard/AutomaticPaymentList.tsx
```diff
@@ -2,12 +2,19 @@ import  React  from 'react';
 import {constants} from "../Resources/Constants";
 import './ViewAutomaticPayments.css'
 import ActivatePaymentToggle from "../AdminDashboard/ActivatePaymentToggle"
+import {CreateAutomaticPaymentForm} from "./CreateAutomaticPaymentModal"
+import add_payment from "../Resources/images/add_payment.png"
```
Slitthe: Please keep this variable's naming style consistent to the rest of the application, which uses `camelCase`, also I think that naming this `addPaymentImg` would be more descriptive

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
```diff
@@ -0,0 +1,258 @@
+import React from "react";
+import { constants } from "../Resources/Constants.js";
+import '../AdminDashboard/AttachCardModal.css';
+
+let regexCheckIfPositiveFloat = "^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$";
+let currentDate = new Date();
```
Slitthe: If you don't change these variables, make them constants `const`

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+        else{
+            fetch(constants.baseUrl+"company", {
+                method: "GET",
+                headers: {
+                'Accept': 'application/json',
+                'Content-Type': 'application/json',
+                'x-access-token' : token.toString()
+                }})
+            .then( response => 
+                {
```
Slitthe, line 155: JS convention: curly bracket on the same line
`blah {`

`}`

ClientApp/viking-vault/src/UserDashboard/ViewAutomaticPayments.css
```diff
@@ -11,4 +11,24 @@
 .automatic-payments-list {
     position: relative;
     right: 4%;
-}
\ No newline at end of file
+}
+
+.add_payment_icon {
```
Slitthe: Keep the class naming convention consistent, your application uses `kebab-case` for CSS naming, use that here as well

Server/VikingVault.Services/AutomaticPaymentService.cs
```diff
@@ -54,5 +55,29 @@ public List<AutomaticPaymentDTO> GetAllAutomaticPayments(string token)
 
             return automaticPaymentData;
         }
+
+        public AutomaticPayment CreateAutomaticPayment(AutomaticPaymentDTO automaticPaymentDTO, string token)
+        {
+            var tokenObject = new JwtSecurityToken(token);
+            int userId = Int32.Parse(tokenObject.Payload["Id"].ToString());
```
Slitthe: Use the same data type in the parse as it is in the variable declaration. (I know that `int` and `Int32` are the same but its confusing)

ClientApp/viking-vault/src/UserDashboard/AutomaticPaymentList.tsx
```diff
@@ -1,10 +1,16 @@
...
 const baseUrl = constants.baseUrl + "AutomaticPayment";
 
+interface IModals {
+    openCreateAutomaticPaymentModal: boolean
+}
+
 interface IAutomaticPayment {
+    companyId: number,
     companyName: string,
     amount: Number,
```
OlahCatalina: Please use "number" for amount instead of "Number" and eventually investigate the difference between these two.

OlahCatalina (approved):

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    private initialPaymentDateIsAfterCurrentDate = () => {
+        if(this.state.automaticPayment.initialPaymentDate !== "" && new Date(this.state.automaticPayment.initialPaymentDate) >= currentDate)
+            return  {
+                message: "Ok",
+                class: "alert alert-success"
+            }
+        return  {
```
crisfsabau, line 93: you have 2 returns in this if, shouldn't the second return be on the else branch?

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+            companyId: this.getIdOfCompanyByName(this.state.automaticPayment.companyName),
+            companyName: this.state.automaticPayment.companyName,
+            amount: this.state.automaticPayment.amount,
+            initialPaymentDate: new Date(this.state.automaticPayment.initialPaymentDate),
+            lastPaymentDate: new Date(2000,0,2),
+        };
+    }
+
+    getAllCompanies = () => {
+        var companies;
```
crisfsabau: don't use var. Use let or const. let for variables that can be reassigned and const for read only variables

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+        var token = sessionStorage.getItem("Authentication-Token");
```
crisfsabau: same as above

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
crisfsabau: is this resolved? I don't see it as resolved

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    private amountMustBePositive = () => {
+        if(this.state.automaticPayment.amount.match(regexCheckIfPositiveFloat))
+            return  {
+                message: "Ok",
+                class: "alert alert-success"
+            }
+        return  {
```
crisfsabau, line 80: Here you have 2 returns in the same branch of the if statement. Did you wanna return after the if, maybe? or in the else branch?

Server/VikingVault.Services/CompanyService.cs
```diff
@@ -73,9 +73,9 @@ public List<CompanyDataDTO> GetAllCompanies()
                 var companies = _dbContext.User.Where(user => user.Role.Type == RoleEnum.Company.ToString()).ToList();
                 foreach (var company in companies)
                 {
-                    var account = _dbContext.BankAccount
+                    var account = _dbContext.BankAccount.Include(bankAccount => bankAccount.User)
                         .SingleOrDefault(bankAccount => bankAccount.User.Id == company.Id &&
-                                                        bankAccount.CurrencyType == CurrencyEnum.Ron.ToString());
+                                                        bankAccount.CurrencyType.Equals("Ron"));
```
crisfsabau, line 78: why did you change this?

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
ed2298: Not necessarily. If cond=true, then the return inside the if statement is returned, which means that the latter return won't be reached. Otherwise, the code inside the if statement is not executed, so the second return is "executed".

crisfsabau: yes but I don't see the closing brace of the if before the return

Server/VikingVault.Services/CompanyService.cs
ed2298: I think it is safer to use equals for strings, instead of == operator.

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
ed2298: That is because there is no opening brace ;)

crisfsabau: Then use braces, to be more clear

ed2298: I will enclose the code from the if statement in curly brackets

PR #58, opened 2019-08-26
Int2019 52 user delete automatic payments

vladbuda:


ClientApp/viking-vault/src/Common/SideBar.tsx
```diff
@@ -34,6 +34,8 @@ class SideBar extends React.Component<ISideBarProps, ISideBarState> {
 
     render(){
         const sidebarVisibility:string = this.state.show ? "sidebar collapsed" : "sidebar expanded";
+        const sideMenuVisibility:string = this.state.show ? "side-menu-hide" : "side-menu-show";
+        const spanVisibility: string = this.state.show ? "span-hide" : "span-show";
```
vasi-axinte: spanVisibility is unclear. Which is the span you are referring to?

ClientApp/viking-vault/src/UserDashboard/DeleteAutomaticPayment.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import deleteIcon from "../Resources/images/delete.png";
+import {constants} from "../Resources/Constants";
+import "./ViewAutomaticPayments.css";
+
+const url = constants.baseUrl+"automaticPayment/";
```
vasi-axinte: URL is not suggestive enough. Try finding something more meaningful.

ClientApp/viking-vault/src/UserDashboard/DeleteAutomaticPayment.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+interface IDeleteAutomaticPaymentProps {
+    automaticPaymentId: number;
+    deletePaymentFromList: (id: number) => void;
+}
+
+function deleteAutomaticPaymentFromDb(automaticPaymentId: number, props: IDeleteAutomaticPaymentProps) {
```
vasi-axinte: I would use database instead of db... Also, how it is possible to "delete from database" directly from your client app? How do you access the database?

ClientApp/viking-vault/src/UserDashboard/DeleteAutomaticPayment.tsx
```diff
@@ -0,0 +1,39 @@
...
+            id: automaticPaymentId
+        })
+    }).then(response => {
+        if(response.status === 200) {
+            props.deletePaymentFromList(automaticPaymentId);
+        }
+    });
+}
+
+function DeleteAutomaticPayment (props: IDeleteAutomaticPaymentProps) {
```
vasi-axinte, line 35: DeleteAutomaticPayment is C# style. Js uses different naming style for functions :)

vasi-axinte (changes requested):

ClientApp/viking-vault/src/Common/SideBar.tsx
```diff
@@ -2,21 +2,21 @@ import React from 'react'
 import './styles.css'
 import dashboard from '../Resources/images/dashboard.png'
 import arrow from '../Resources/images/arrowGRIdeschis.png'
+import placeholder from '../Resources/images/placeholder.png'
+import paymentMethod from '../Resources/images/payment-method.png'
+import block from '../Resources/images/password.png'
```
Slitthe: Wouldn't really hurt to name imports related to the images be more descriptive, when I see the name `block` in the code I don't know what it refers to *Passive aggressive look*

ClientApp/viking-vault/src/UserDashboard/AutomaticPaymentList.tsx
```diff
@@ -90,7 +101,7 @@ class AutomaticPaymentList extends React.Component<any, IAutomaticPaymentsState>
                 <td className="payments-text centered-text">{this.formatDate(payment.lastPaymentDate)}</td>
                 <td className="payments-text centered-text"><ActivatePaymentToggle paymentId={payment.id} /></td>
                 <td className="payments-text centered-text">Edit element here</td>
-                <td className="payments-text centered-text">Delete element here</td>
+                <td><DeleteAutomaticPayment automaticPaymentId = {payment.id} deletePaymentFromList = {this.deletePaymentFromList}/></td>
```
Slitthe: Don't make the code hard to read just because you want to keep things in one line, in this case I think it will be better from a code readability standpoint if you'd move `<DeleteAutomaticPayment` on its own line

Server/VikingVault.API/Controllers/AutomaticPaymentController.cs
```diff
@@ -19,12 +21,27 @@ public AutomaticPaymentController(IAutomaticPaymentService automaticPaymentServi
...
         }
+
+        [HttpDelete]
+        [Authorization(Role = RoleEnum.User)]
+        public ActionResult Delete(AutomaticPaymentId automaticPaymentToDelete)
+        {
+            try
+            {
+                _automaticPaymentService.DeleteAutomaticPayment(automaticPaymentToDelete.Id);
+                return Ok(true);
```
Slitthe: Why return `true` if it's never otherwise?

ClientApp/viking-vault/src/UserDashboard/AutomaticPaymentList.tsx
```diff
@@ -81,6 +82,16 @@ class AutomaticPaymentList extends React.Component<any, IAutomaticPaymentsState>
         this.getAutomaticPayments();
     }
 
+    deletePaymentFromList = (id : number) =>{
+        let paymentList = this.state.payments.filter(payment => {
```
Slitthe: Opposite in here, sometimes It makes more sense to use one liner's for simple methods such as `.filter()` and single `setState`. 

Just try think at all times that a new person might join your team and would have to read your code and put yourself in their shoes, how would would you want the code to be written in such a way that it would make their life easier?...

`let paymentList = this.state.payments.filter(payment => payment.id !== id)`

`this.setState({ payments : paymentList });`

ClientApp/viking-vault/src/UserDashboard/DeleteAutomaticPayment.tsx
vladbuda: That's the name of a React component (functional), therefore I think it should start with an uppercase.

ClientApp/viking-vault/src/Common/SideBar.tsx
vladbuda: That was a bad import name, indeed, but we no longer need it, therefore it was deleted.

Server/VikingVault.API/Controllers/AutomaticPaymentController.cs
vladbuda: An internal server error can also be returned in case of an exception from UserService.

Server/VikingVault.API/Startup.cs
```diff
@@ -69,7 +69,7 @@ public void ConfigureServices(IServiceCollection services)
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IHostingEnvironment env, VikingVaultDbContext context)
         {
-	        context.Database.Migrate();
+	        //context.Database.Migrate();
```
Slitthe: Delete this line of commented code if you don't need it

vladbuda: That line shouldn't be commented. I removed the comment.

PR #61, opened 2019-08-27
ASAP:Dashboard icon redirects to dashboard

OlahCatalina:
+ other titles for icons


ClientApp/viking-vault/src/Common/UserIcon.tsx
```diff
@@ -100,7 +100,7 @@ class UserIcon extends React.Component<any, IUserIconState> {
 
   render() {
     return (
-      <div className="dropdown" ref={this.megaMenu}>
+      <div title="Profile" className="dropdown" ref={this.megaMenu}>
```
vasi-axinte: What is megaMenu?

vasi-axinte (changes requested):

ClientApp/viking-vault/src/Common/UserIcon.tsx
OlahCatalina: The menu with the **_"View Profile"_** and **_"Logout"_** options.

vasi-axinte: It`s confusing, so it should be changed to something meaningful.

PR #62, opened 2019-08-27
ASAP:Int2019-added notifications to database

moleguaco99:


Server/VikingVault.Services/ScopedProcessingService.cs
```diff
@@ -46,6 +46,7 @@ public void PayCompanies()
                     automaticPayment.IsEnabled = false;
                     _dbContext.AutomaticPayments.Update(automaticPayment);
                     _dbContext.SaveChanges();
+                    AddFailedPaymentNotification(automaticPayment);
```
vasi-axinte: AddFailedPaymentNotification if an update is succesful? It looks weird to me.

vasi-axinte (changes requested):

Server/VikingVault.Services/ScopedProcessingService.cs
moleguaco99: How do you suggest to change that, considering that the update is made at first, so that the automatic payment will be disabled, and then we notify the user that the payment failed? Should I refactor the update in a separate function?

Server/VikingVault.Services/TransferRequestService.cs
```diff
@@ -23,12 +27,27 @@ public TransferRequest AddTransferRequest(TransferRequest transferRequestData)
...
             catch (Exception e)
             {
                 throw new DatabaseException();
             }
             return transferRequestData;
         }
+
+        private void AddRequestNotification(TransferRequest transferRequestData)
+        {
+            int? idReceiver = _userCardService.FindUserIdByCardNumber(transferRequestData.CardNumberReciever);
```
Slitthe, line 43: You make this a nullable int but then you don't check for null values.... *judging intensifies*
`(int)null` will not work, just saying

Server/VikingVault.Services/TransferFundsService.cs
```diff
@@ -38,11 +38,25 @@ public void TransferFunds(TransferFundsModel transferData)
...
-
-                _transactionService.AddTransactionsForTransferFunds(transferData);          
+                _transactionService.AddTransactionsForTransferFunds(transferData);
+                AddReceivedNotification(transferData, idReciever);
             }
         }
 
+        private void AddReceivedNotification(TransferFundsModel transferData, int? idReciever)
+        {
+            User receiver = _userService.GetById((int)idReciever);
```
Slitthe: Same as below 

*more judgement*

Server/VikingVault.Services/ExecutePaymentService.cs
```diff
@@ -22,44 +22,77 @@ public ExecutePaymentService(VikingVaultDbContext dbContext)
...
+            _dbContext.AutomaticPayments.Update(automaticPayment);
+            _dbContext.SaveChanges();
+        }
+
+        private void AddSuccessfulPaymentNotification(AutomaticPayment automaticPayment)
+        {
+            Notification notification = new Notification
+            {
+                User = automaticPayment.PayingUser,
+                Text = "Payment to " + automaticPayment.ReceivingCompany.FirstName +
```
Slitthe: I think this will look better using string interpolation `$"Basic chars here {awesomeInterpolatedVariableHere}"`

Server/VikingVault.Services/TransferFundsService.cs
```diff
@@ -38,11 +38,25 @@ public void TransferFunds(TransferFundsModel transferData)
             {
                 _bankAccountService.RetractMoneyFromUser(_userService.GetById(idSender), transferData.Currency, transferData.AmountSent);
                 _bankAccountService.AddMoneyToUser(_userService.GetById((int)idReciever), transferData.Currency, transferData.AmountSent);
```
Slitthe: Null checks yo! (see below)

Server/VikingVault.Services/TransferRequestService.cs
moleguaco99: Improvise.Adapt.Overcome

moleguaco99: I checked it

PR #70, opened 2019-08-28
ASAP: Int2019 73 fix admin dashboard delete user

vladbuda:


ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -30,6 +41,20 @@ class TransactionList extends React.Component<any, IState> {
...
+            if(transaction.receiver === null){
+                transaction.receiver = deletedUser;
+            }
+            return transaction;
+        });
+        return transactions;
+    }
+
     private getTransactions() {
         let token = sessionStorage.getItem("Authentication-Token");
```
crisfsabau: token should be declared with const, because as I see you don't change it

Server/VikingVault.Services/UserService.cs
```diff
@@ -105,6 +105,9 @@ public void DeleteUser(UserEmail userEmail)
             try
             {
                 User userToDelete = _dbContext.User.SingleOrDefault(user => user.Email == userEmail.Email);
+                _dbContext.Transactions.Where(transaction => transaction.Sender.Id == userToDelete.Id).Load();
+                _dbContext.Transactions.Where(transaction => transaction.Receiver.Id == userToDelete.Id).Load();
+
                 if(userToDelete == null)
```
crisfsabau: You should check for null before you get the transactions, because if it's null there is no point to check or his transactions

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -30,6 +41,20 @@ class TransactionList extends React.Component<any, IState> {
         transactions: []
     }
 
+    private checkIfUserNull(result:ITransaction[]) {
```
crisfsabau: A function named checkIfUserNull is telling me that it should return a true/false value. Actually, I see that t returns a list of transactions. Rename the user accordingly and also, specify the return type of the function.

crisfsabau (changes requested):

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -30,8 +41,22 @@ class TransactionList extends React.Component<any, IState> {
         transactions: []
     }
 
+    private changeNullUsersToDeleted(result:ITransaction[]): ITransaction[] {
```
crisfsabau: changeNullUsersToBeDeleted maybe :)

vladbuda: Null users are not deleted, just replaced by a mocked user

ClientApp/viking-vault/src/UserDashboard/TransactionList.tsx
```diff
@@ -21,6 +21,17 @@ interface ITransaction {
...
 
+const deletedUser : IUserData = {
+    id: 0,
+    firstName: "DeletedUser",
+    lastName: "",
+    address: "DeletedUser",
+    email: "DeletedUser",
+    pictureLink: "DeletedUser",
+    cardNumber: "DeletedUser",
+    expirationDate: "DeletedUser"
```
crisfsabau: Don't initialize every property with deletedUser

PR #72, opened 2019-08-28
Int2019 - Fixed PDF

OlahCatalina:


Server/VikingVault.Services/PDFGeneratorService.cs
```diff
@@ -81,30 +81,65 @@ public void AddTransactionListIntoPDF(PdfPage page, string userId, string timeFi
...
+                    graphics.DrawString(transactionDetails, columnDataFont, columnDataBrush, new PointF(x + 390, y));
+                }
                 y = y + 25;
             }
         }
 
+        public string TrimmTransactionDetails (TransactionDTO transaction)
+        {
+            string transactionDetails = transaction.Details.ToString();
+            if (transactionDetails.Length > 23)
```
crisfsabau: You can extract 23 in a constant, as it used also below

Server/VikingVault.Services/PDFGeneratorService.cs
```diff
@@ -81,30 +81,65 @@ public void AddTransactionListIntoPDF(PdfPage page, string userId, string timeFi
...
+                graphics.DrawString(transaction.Currency.ToUpper(), columnDataFont, columnDataBrush, new PointF(x + 180, y));
+
+                if (transaction.Type == "Exchange")
+                {
+                    graphics.DrawString("Exchanged into " + transaction.Details.ToString(), columnDataFont, columnDataBrush, new PointF(x + 240, y));
+                    graphics.DrawString("Money Exchange", columnDataFont, columnDataBrush, new PointF(x + 390, y));
+                }
+                else
+                {
+                    string transferText = "";
```
crisfsabau: instead of "" use string.Empty, because "" creates a new object, so from the memory efficiency point of view string.Empty is the winner.

Server/VikingVault.Services/PDFGeneratorService.cs
```diff
@@ -81,30 +81,65 @@ public void AddTransactionListIntoPDF(PdfPage page, string userId, string timeFi
...
+                    graphics.DrawString(transferText, columnDataFont, columnDataBrush, new PointF(x + 240, y));
+                    graphics.DrawString(transactionDetails, columnDataFont, columnDataBrush, new PointF(x + 390, y));
+                }
                 y = y + 25;
             }
         }
 
+        public string TrimmTransactionDetails (TransactionDTO transaction)
+        {
+            string transactionDetails = transaction.Details.ToString();
```
crisfsabau: I would use **var** any time the initialization of the variable clearly tells the type. For Example here you call ToString() method, it is clear that this is a string

crisfsabau (changes requested):

Server/VikingVault.Services/PDFGeneratorService.cs
OlahCatalina: Done, thank you!

Server/VikingVault.Services/PDFGeneratorService.cs
OlahCatalina: Uuu, didn't know about this. Thank you, I changed it.

Server/VikingVault.Services/PDFGeneratorService.cs
OlahCatalina: Changed it.

PR #79, opened 2019-08-28
ASAP: INT2019 85 fix refresh automatic payment

ed2298:


ClientApp/viking-vault/src/UserDashboard/ViewAutomaticPayments.tsx
```diff
@@ -6,20 +6,20 @@ import UserIcon from "../Common/UserIcon";
 import { AutomaticPaymentList } from "./AutomaticPaymentList";
 import "../Common/styles.css";
 
-export interface IAutomaticPaymentsState {
-    reload: boolean;
-}
+// export interface IAutomaticPaymentsState {
```
vasi-axinte: Please delete unused code.

vasi-axinte (changes requested):

ClientApp/viking-vault/src/UserDashboard/CreateAutomaticPaymentModal.tsx
```diff
@@ -2,14 +2,14 @@ import React from "react";
 import { constants } from "../Resources/Constants.js";
 import '../AdminDashboard/AttachCardModal.css';
 
-const regexCheckIfPositiveFloat = "^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$";
+const regexCheckIfPositiveFloat = "^[-+]?[0-9]*\.?[0-9]+([-+]?[0-9]+)?$";
```
Slitthe: Please extract this and move into the constants file as it seems like the same regex is used in multiple pages within the app. So when you want to make a change in this regex you then have to modify it in all areas, versus if you were to extract it you would only have to modify it in 1 place

PR #82, opened 2019-08-29
PDF download fixed

OlahCatalina:
created a new method that gets ALL the transactions

### DIFF
--- a/Server/VikingVault.Services/TransactionService.cs
+++ b/Server/VikingVault.Services/TransactionService.cs
@@ -98,5 +98,38 @@ namespace VikingVault.Services
             }
             return transactionsDTO;
         }
+
+        public List<TransactionDTO> GetAllTransactions(string userId)
+        {
+            int uid = int.Parse(userId);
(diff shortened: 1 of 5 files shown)


Server/VikingVault.Services/TransactionService.cs
```diff
@@ -98,5 +98,38 @@ public List<TransactionDTO> GetTransactions(string userId)
             }
             return transactionsDTO;
         }
+
+        public List<TransactionDTO> GetAllTransactions(string userId)
+        {
+            int uid = int.Parse(userId);
```
Slitthe: Please be aware that the `int.Parse` can throw an exception if it cannot parse the string

Slitthe: Please be aware that the `int.Parse` can throw an exception if it cannot parse the given string. (Just in case you want to catch this somewhere I'm saying)

OlahCatalina: Understood, I've changed it. Thank you!

PR #84, opened 2019-08-29
INT2019-ShrinkPaymentToggling

moleguaco99:
-minimized toggle for payment
-grayed-out inactive payment

### DIFF
--- a/ClientApp/viking-vault/src/AdminDashboard/ActivatePaymentToggle.tsx
+++ b/ClientApp/viking-vault/src/AdminDashboard/ActivatePaymentToggle.tsx
@@ -7,7 +7,7 @@ const paymentTogglingURL = `${constants.baseUrl}PaymentToggling/`;
 
 interface IPaymentToggleProps{
   paymentId: number
-  toggleAutomaticPaymentState: (id: number, state:boolean) => void
+  toggleAutomaticPaymentState: (id: number, automaticPaymentState:boolean) => void
 }
 
 class ActivatePaymentToggle extends React.Component<IPaymentToggleProps, any>{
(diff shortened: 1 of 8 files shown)


ClientApp/viking-vault/src/AdminDashboard/ActivatePaymentToggle.tsx
```diff
@@ -7,6 +7,7 @@ const paymentTogglingURL = `${constants.baseUrl}PaymentToggling/`;
 
 interface IPaymentToggleProps{
   paymentId: number
+  toggleAutomaticPaymentState: (id: number, state:boolean) => void
```
Slitthe: `state` of what? More description please

ClientApp/viking-vault/src/AdminDashboard/toggle.css
```diff
@@ -56,4 +56,8 @@
   
   .slider.round:before {
     border-radius: 50%;
+  }
+
+  .scaled{
```
Slitthe: Maybe `scaled-down` is better, as `scaled` could mean both scaled up or down

ClientApp/viking-vault/src/UserDashboard/AutomaticPaymentList.tsx
```diff
@@ -102,16 +103,30 @@ class AutomaticPaymentList extends React.Component<any, IAutomaticPaymentsState>
         this.setState({payments : paymentList});
     }
 
+    toggleAutomaticPaymentState = (id: number, state:boolean) =>{
+        let automaticPayments = this.state.payments.map((automaticPayment:IAutomaticPayment) => {
```
Slitthe: You could use the array function `.forEach` in here.

blah.forEach(item => {
   if(blah) {
       item.something = somethingElse;
   }
})

ClientApp/viking-vault/src/UserDashboard/AutomaticPaymentList.tsx
```diff
@@ -102,16 +103,30 @@ class AutomaticPaymentList extends React.Component<any, IAutomaticPaymentsState>
         this.setState({payments : paymentList});
     }
 
+    toggleAutomaticPaymentState = (id: number, state:boolean) =>{
```
Slitthe: Since you marked the function in the interface as being `void`, make it `void` here as well

